PR #222, opened 2014-08-17
Fix can't reference main bundles in production mode uniquely

asavoy:
This fix actually makes the problem more apparent: if steal is asked to load a main bundle in production, it will only use the filename component. So `users/list.js` and `tasks/list.js` would both result in looking for `bundles/list.js`. Instead, this keeps the path component, and so the mains will be unique: `users/list.js` => `bundles/users/list.js` and `tasks/list.js` => `bundles/tasks/list.js`. 

This is the corresponding fix to https://github.com/bitovi/steal-tools/pull/54 - that fix is required to have steal-tools build main bundles at the location that this fix expects them to be.


justinbmeyer:
@asavoy I'm sorry to even ask this, but if you can rebase this one more time, I can get this in on Thursday for the 0.6.0 release.  Thanks!

alexisabril:
Rebased version of this PR is in #320

PR #265, opened 2014-10-10
Less options paths

whitecolor:
Added ability to have custom ```paths``` option in ```lessOptions``` (that is used only in nodejs mode). Options moved inside of ```translate``` function. Added ```steal.extend```.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -15,7 +15,9 @@
     "notStrictEqual": true,
     "raises": true,
     "start": true,
-    "stop": true
+    "stop": true,
+    "System": true,
+    "Promise": true
   },
 
   "curly": true,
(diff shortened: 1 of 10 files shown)


justinbmeyer:
@whitecolor Sorry about the delay.  I'm trying to get everything in for 0.6.0. Can you:

- remove the steal.extend, 
- add docs
- change the test to use module value of the less file.  It is and object with a `source` property that is the compiled CSS output.  https://github.com/bitovi/steal/blob/master/ext/css.js#L39

If you can do this, I will merge it.  Thanks!

whitecolor:
@justinbmeyer I would better move less support out of steal core. I currently use my own plugin. I'm not gotint to update this pull so I think it should be closed.

justinbmeyer:
One of the central motivations of steal is to make it so less and a few other types work by default.

PR #298, opened 2014-12-11
Allow truthy script options.

### LINKED ISSUE #288: Allow truthy script options
For example:

```html
<script src="bower_components/steal/steal.js"
  data-bower-dev></script>
```

Should make `System.bowerDev === true`.  I think this doesn't work because we are just setting to the attribute value.

m-mujica:
Closes #288.

As per [spec](http://www.w3.org/TR/html-markup/syntax.html#syntax-attr-empty), the empty attribute syntax `<input disabled>` is exactly equivalent to specifying the empty string as the value for the attribute, as in `<input disabled="">`, I tried it in a few browsers and the behaviour is consist... I'll try later tonight in IE.


test/test.js
```diff
@@ -243,6 +243,10 @@ QUnit.config.testTimeout = 30000;
 		makeIframe("less_options/site.html");
 	});
 
+	asyncTest("allow truthy script options", function(){
```
justinbmeyer: include the issue number

PR #313, opened 2015-02-03
Moved Quick Start to separate file and update stealjs overview page.

cklanac:
Moved Quick Start to a separate file and removed the content from stealjs.md. And added Quick Start to the Guides navigation as the second item after Why StealJS. 

Also updated several details on Quick Start

### DIFF
--- a/docs/pages/quick-start.md
+++ b/docs/pages/quick-start.md
@@ -1,0 +1,132 @@
+@page StealJS.quick-start Quick Start
+@parent StealJS.guides 1
+
+## Quick Start
+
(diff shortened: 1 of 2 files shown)


justinbmeyer:
@cklanac Why did you move it?  Normally, we have a quick start on the overview page.  People actually seem to prefer this instead of having to click on a lot of pages to get started.

cklanac:
Since it isn't referenced in the Guides navigation it's a bit hard to find, particularly if someone hasn't entered from the main page. 

If you prefer, I can recombine them. But I think the Overview or QuickStart should be added to the Guides navigation.

justinbmeyer:
Hmmm, that's a valid point.

PR #329, opened 2015-02-19
Windows Compatibility

James0x57:
Adding windows compatibility - Please review =)


justinbmeyer:
Please add testee back in for now so this doesn't get merged without it.

James0x57:
$ npm install
and
$ grunt build
both went through without error after re-adding testee.

matthewp:
Ok, I'll try running this through the Windows CI (appveyor)

matthewp:
It's failing because we don't have browsers set up in Appveyor, i'm looking up how to do this.

main.js
```diff
@@ -42,6 +42,8 @@
 		},
 		dir = function(uri){
 			var lastSlash = uri.lastIndexOf("/");
+			if(lastSlash === -1)
```
matthewp: Add a comment for why we are doing this.

main.js
```diff
@@ -299,6 +303,24 @@ var makeSteal = function(System){
 				addProductionBundles.call(this);
 			}
 		};
+
+	// checks if we're running in node, then prepends the "file:" protocol if we are
+	var envPath = function(val) {
+		//next line is var we can use to check if it's running on windows. May use if specifying the "file:" protocol causes issues on *nix machines.
+		//var isInWindows = (typeof process !== "undefined" && typeof process.platform !== "undefined" && /^win/.test(process.platform));
```
matthewp: Get rid of this code, we know this works fine on unix.

matthewp:
We should do a setter for `paths` as well.

matthewp:
Yay, finally. There is a bug in npm v1.x that causes npm install to break on Windows some times, but happened often on AppVeyor. Upgrading to npm 2.x on AppVeyor fixes the issue and now we have a clean build.

matthewp:
Oh, we need to get a path setter working with this as well. I'll update it.

PR #330, opened 2015-02-20
IE8 doesn't support console.log.apply

tracer99:


ext/dev.js
```diff
@@ -117,10 +117,14 @@
 			}
 		},
 		_logger:function(type, arr){
-			if(console.log.apply){
-				console[type].apply(console, arr);
-			} else {
+			if (typeof console.log == "object") {
```
matthewp: What is this case for?

tracer99: This is feature detection, specifically for IE8. console.log is an object in IE8 but a function in all other modern browsers. console.log.apply does not exist in IE8

matthewp: If it's an object then how do you use it? Won't `console[type](arr);` throw?

tracer99: Nope. Just tried it in IE8.

matthewp: Can you explain though?  If `type` is "log" and you are saying that console.log is an object, you can't call an object, it should throw. I don't doubt that your code is correct, I just want to understand it.

matthewp: Ok, I researched this a bit myself and it seems that console.log is an object but you can call it like a function. Really bizarre.  However I also found out that `console` doesn't exist until you open the dev tools. So we should really be checking typeof console as well.

tracer99: Trying to explain how IE8 works half of the time is an exercise in frustration.

tracer99: Sure. I'll amend the code

matthewp:
Thank you!

PR #341, opened 2015-02-26
Fixing links

marshallswain:
This fixes the link to npm/http-server and turns the URLs to localhost:8080 into links.  They might as well be clickable.


matthewp:
Awesome, thank you.

Issue #372, opened 2015-04-06
Less options paths again


whitecolor:
Currenlty `paths` in `lessOptions` is ignored, I think optional paths should be allowed to be added. 
```
lessOptions{
  path: [__dirname]
},
```
and this value should be concatinated with internal paths.

matthewp:
Closed by https://github.com/stealjs/steal/pull/378

PR #373, opened 2015-04-06
Allow additional paths in lessOptions

whitecolor:
https://github.com/stealjs/steal/issues/372

### DIFF
--- a/docs/system-lessOptions.md
+++ b/docs/system-lessOptions.md
@@ -42,5 +42,7 @@ will be compiled to:
 }
 ```
 
-`lessOptions.paths` and `lessOptions.filename` are used internally by StealJS and any value
+Use can also provide `lessOptions.paths` (array) option which will be used to resolve imports (see less/css docs).
+
+`lessOptions.filename` are used internally by StealJS and any value
 provided will be ignored.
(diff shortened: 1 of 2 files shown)


matthewp:
Thanks for submitting. Can you give an example of what this would allow you to do that you couldn't previously do? I think it's a fair change, just want to understand what this allows now.

whitecolor:
In my less files I use absolute paths

Say I have `styles` directory where I put common imports files for my project, and then in any file I just include:

```less
@import "/styles/some/common.less"
```

So one maysay: use relative, it is better not to use absolute paths and so on... but currently it is more appropriate for my flow to use such paths. I dont have to care about where my particular file (with `@import` statement) is located. I just put absolute path there. and it works fine. if it would be absolutly no need in this option for user, it would be probably removed from `less/css` compiler. So I think you should allow users to define it. And even more I would include `baseURL` dir in `paths` by default, for convinice, I will break nothing but will demand less config from user the whant to use `absolute` urls (wich are actually relative to baseURL dir).

matthewp:
Thanks for the explanation, will review the code and issue tomorrow (been a while since I thought about less path rules).

matthewp:
Looks good to me, if you can update the PR towards the `minor` branch I'll merge it.  It can go in 0.9.0 as a new feature.

matthewp:
Merged in minor in #378

PR #378, opened 2015-04-08
Add support for lessOptions.paths option

whitecolor:
Add support for lessOptions.paths option and docs updated

### DIFF
--- a/docs/system-lessOptions.md
+++ b/docs/system-lessOptions.md
@@ -42,5 +42,7 @@ will be compiled to:
 }
 ```
 
-`lessOptions.paths` and `lessOptions.filename` are used internally by StealJS and any value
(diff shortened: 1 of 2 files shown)


docs/system-lessOptions.md
```diff
@@ -42,5 +42,7 @@ will be compiled to:
 }
 ```
 
-`lessOptions.paths` and `lessOptions.filename` are used internally by StealJS and any value
+Use can also provide `lessOptions.paths` (array) option which will be used to resolve imports (see less/css docs).
```
matthewp: I think you mean `You can` here.  Otherwise looks good, will merge.

matthewp:
Sorry for taking so long to get to this, just one grammatically error in the docs and then I can merge it.

matthewp:
Thanks!

Issue #384, opened 2015-04-10
Webworkers and StealJS


justinbmeyer:
I would like to be able to use StealJS in a webworker scenario. 

I'd like to have steal import a module in one step like:

```
var worker = new Worker(System.stealPath+"?main=webworkers/worker");
module.exports = worker;
```

This means making 2 changes:

1.  Provide a `System.stealPath` that represents where the `steal.js` file is located.  We already set this, but the value doesn't show up on `System`
2.  If steal.js is loaded in a webworker, it should parse the [location](https://developer.mozilla.org/en-US/docs/Web/API/Location)'s `search` for configuration options.

I think this is an easy change, but will make it possible to quickly load modules in a worker-thread.

BigAB:
:+1:

matthewp:
Closed by #386

justinbmeyer:
@matthewp Should this be documented?  Maybe reopen until then?

matthewp:
Sure, where would you document this?

matthewp:
This just needs to be documented now, free for anyone who wants to do it.

justinbmeyer:
open for docs, but un assigning myself

matthewp:
I'm not a fan of this API and think we should hold off documenting until we can find a better one.  It will continue to work this way, of course.

PR #386, opened 2015-04-11
makes webworkers work with a test

justinbmeyer:
For #384.

Works like:

```js
var worker = new Worker(System.stealURL+"?main=webworkers/worker&config="+System.configPath);
```

You point at steal, pass the configPath, main, and any other options needed by steal.

Other thoughts:

 - should we change configPath to configURL? This would make baseURL, stealURL, configURL all consistent.


src/start.js
```diff
@@ -93,4 +93,7 @@
 				removeDotSegments(href.protocol || href.authority || href.pathname.charAt(0) === '/' ? href.pathname : (href.pathname ? ((base.authority && !base.pathname ? '/' : '') + base.pathname.slice(0, base.pathname.lastIndexOf('/') + 1) + href.pathname) : base.pathname)) +
 					(href.protocol || href.authority || href.pathname ? href.search : (href.search || base.search)) +
 					href.hash;
-		};
\ No newline at end of file
+		},
+		isWebWorker = typeof WorkerGlobalScope !== 'undefined' && self instanceof WorkerGlobalScope,
+		isBrowserWithWindow = typeof window === "undefined",
```
matthewp: This will be false in the browser. I think you wanted `!==`

src/config.js
```diff
@@ -2,6 +2,8 @@
 	var setterConfig = function(loader, configSpecial){
 		var oldConfig = loader.config;
 		
+		var tval = loader.paths["@traceur"];
```
matthewp: What is this for?

PR #409, opened 2015-04-24
Add live-reload extension

matthewp:
This adds the new live-reload extension. Sets the path for `live-reload` to `ext/live-reload.js` so this can be used automatically as a configDependencies:

```json
{
  "system": {
    "configDependencies": [
      "live-reload"
    ]
  }
}
```


matthewp:
Whoops, need to update the docs.

PR #432, opened 2015-06-01
Add steal-css dependency.

m-mujica:
The css plugin now lives in its own repo https://github.com/stealjs/steal-css.


matthewp:
Can you remove the `asset-register` stuff? That's now implemented by another plugin.

PR #442, opened 2015-07-01
@steal is a module of the steal loading

### LINKED ISSUE #424: Add a @steal module
This would like `@loader`, it refers to the steal that is loading the given module.

matthewp:
This allows you to do:

```js
steal("@steal", function(steal){
  // I can use steal.import, etc
});

var steal = require("@steal");
```

To use the functionality that the steal object provides. Closes #424


matthewp:
This needs docs before merge.

PR #488, opened 2015-09-29
Get all tests to pass in IE8

matthewp:
This fixes our IE8 compatibility. There were a few uses of `loader.import` rather
than `loader["import"]`.


matthewp:
screen shot 2015-09-29 at 10 02 16 am

PR #504, opened 2015-10-19
Make the bang optional when using System.ext

### LINKED ISSUE #503: Make ext extension work without the !
I think the problem that causes this not to work is that the plugin extension will individually normalize both sides of a module name.  So if our extension makes `foo.css` become `foo.css!css-plugin` then there will be a cycle where it continuously does this over and over.  I think we have to flag this some how in the plugin extension so that the normalize hook knows what is happening.

matthewp:
This makes the use of the `!` option when using System.ext. This means you can now do:

```js
import "foo.css";
```

If you have configured a css ext (which comes by default with StealJS. This is possible because of a change in our SystemJS fork (https://github.com/stealjs/systemjs/commit/f6d7a2174cd2af6fcafb6a3f68066e3cd8f760c5) now passes a flag when performing normalize on the plugin argument, so there is no infinite recursion problem.

Fixes #503


marshallswain:
🙌Glory hallelujah! 🙌

jasongrout:
+1. Thanks!

PR #507, opened 2015-10-27
Add special @less-engine module.

m-mujica:


m-mujica:
@matthewp in `steal-less`, besides changing `less.js` to require `@less-engine` instead of `less`; what should we do while the changes get to steal? I can change the require, but it won't work unless we update its local less. (maybe adding `@less-engine` to `steal-less`'s paths?)

matthewp:
@m-mujica I would add the new steal-less to this PR. So it all goes in at once.

Also, can you get rid of  da77bae? Don't need to push these changes, they happen usually because you have a different version of some plugin than the latest.  The only one I would add is the steal.js and steal.production.js.

m-mujica:
@matthewp done! I removed the extra files, released `steal-less@0.0.2` and bumped the version in `steal`'s package.json.

matthewp:
We'll get this in 0.13.

nlundquist:
Any chance we could get this merged? Current steal-less is incompatible with current steal until we do.

matthewp:
Yep, if you can rebase this i'll merge it.

PR #513, opened 2015-11-16
Use SystemJS as an HTTPS dependency

daffl:
NPM does not allow aliasing module names (so we can't just drop in a published `steal-systemjs` without having to change all references). Making it an HTTPS dependency should still solve the Git problem and is easier to handle.


matthewp:
I still think we should publish this one, there shouldn't be too many references that need to be updated.

daffl:
Is there really a difference except for Semver? As long as we are using `.bitovi.<x>` version numbers we'll have to update the dependency manually every time anyway.

matthewp:
I'm not sure, I thought git dependencies downloaded new every time you did an npm install because they had no way to know that it is unchanged. This isn't a git dependency but I would imagine it works the same way, at any time you can force push to a tag.

Is there a downside to publishing? [This will need to change](https://github.com/stealjs/steal/blob/master/src/end.js#L2) to `steal-systemjs` and the Gruntfile updated but other than that I think there won't be too many other changes.

daffl:
I found the bigger problem to be the `es6-module-loader` references in SystemJS. It looked like it may become a pain merging SystemJS upstream back in.

matthewp:
How many are there?  I think there's this: https://github.com/stealjs/systemjs/blob/master/lib/polyfill-wrapper-end.js#L88  everything in dist/ is built off of that.

daffl:
All done now. Tests should also pass.

PR #532, opened 2015-12-10
Use steal-env

### LINKED ISSUE #531: isEnv, getEnv, etc. are not cloneable

matthewp:
steal-env contains the `getEnv`, `getPlatform`, `isEnv`, `isPlatform` functions that were pulled out of `src/config.js`. These were pulled out so that they are added as part of an extension, so that they work when being cloned.

Fixes #531


matthewp:
Going to resubmit this to a new branch "patch" so that it can be applied to 012.x

Issue #537, opened 2015-12-16
Warn user when they forget the main in production


matthewp:
If you do:

```html
<script src="node_modules/steal/steal.production.js"></script>
```

And forget to include the **main** attribute it will load the dev files. We should warn when you are using `loadBundles` and not providing the main.

matthewp:
Warning would go here: https://github.com/stealjs/steal/blob/master/src/startup.js#L63

It would be something like (pseudo-code):

```js
if(!System.main && System.isEnv("production")) {
  console.warn("A main is needed in production. Please add it to the script tag.");
}
```

PR #540, opened 2015-12-29
warning user when 'main' is not set in production environment

phillipskevin:
Closes https://github.com/stealjs/steal/issues/537.

I'm throwing an error since `console.warn` calls will be removed from `steal.production.js`. Let me know if you want me to handle this some other way.


justinbmeyer:
I don't think we should throw. It's possible that the main might by dynamically imported based on something in the page. 

You can be tricky and do something like:

var warn = console.warn

warn.call(console, ...)

phillipskevin:
@justinbmeyer I updated the PR to do it that way. Thanks.

matthewp:
I think IE8 doesn't support console.warn, or only supports it if the devtools are open, not sure which, should probably account for that.

phillipskevin:
@matthewp IE8 creates the `console` object after devtools are opened. I changed the code to `var warn = console && console.warn || function() {};`. There are a few other `console.log` statements that also won't work in IE8; should I put in a new issue for those?

matthewp:
Yeah, a new issue is a good idea, thanks.

phillipskevin:
https://github.com/stealjs/steal/issues/541 is the new issue.

matthewp:
cool, this will be in the next release, thanks.

PR #561, opened 2016-01-20
Upgrade to Babel 6

### LINKED ISSUE #546: Use Babel 6 by default
There are a few issues that have indirectly requested using babel 6 (https://github.com/stealjs/steal/issues/545, https://github.com/stealjs/steal/issues/535). It looks like Babel is no longer going to have a browser version, so using `babel-core` might be the way to go. Otherwise, it may be necessary to bundle Babel before using it - if that's possible.

matthewp:
This upgrades our Babel to the latest version in babel-standalone. Eventually it would be good to hook up Greenkeeper or something so that we automatically add new patch releases back into Steal.

This also sets Babel to be the default transpiler. Closes #546


matthewp:
Needs documentation.

matthewp:
Blocked by https://phabricator.babeljs.io/T7015

Issue #277, opened 2014-10-31
Support ~ as a way to use locate


matthewp:
For things that do not go through the loader, like less imports, css urls, ~ can be a way to have users opt-in to having certain things use the module loader for locating.

This is tracking a change that will be needed in several places, correct me if I've missed one:

* [x] - Importing javascript: `import foo from "~/foo"`, `System.import("~/foo")`, etc.
* [x] - Import in less: `@import "~/variables.less";`
* [x] - css urls: `url('~/my-font.woff)` (also in less)

There will also be a separate issue in CanJS that allows:

* [x] - Importing partials: `{{> ~/address/address.mustache}}`
* [ ] - Loading images in a `<img>` tag: `<img src="~/images/boat.jpg"/>`

matthewp:
This will probably be implemented as a separate plugin.

matthewp:
We have the `locate://` [scheme now](http://stealjs.com/docs/locate.html).  We still want a way to prevent having to write the package name, creating https://github.com/stealjs/steal/issues/719 for that.

Issue #414, opened 2015-04-28
System.extensions


matthewp:
I'd like to propose a new property on System called `System.extensions`. The problem this is trying to fix is:

1. configDependencies load before the npm plugin, therefore they cannot be an npm dependency. If they are, paths have to be defined prior.
2. configDependencies cannot have npm dependencies because they are loaded before the npm plugin. You can workaround by importing them dynamically after the `configMain` has loaded, but if we want an ecosystem of extensions you want to be able to import your deps statically.

### @property {Array.\<String\>} System.extensions

This would be an array of moduleNames.

My proposal is to `System.import` these modules after `configDeferred` but before `appDeferred`. I also propose that extensions have a signature of:

### function(loader)

Since they will have a single default export this means we can import all extensions at the same time, but execute them in order.

matthewp:
Replaced by #671

PR #565, opened 2016-01-27
Tilde extension to allow replacing plugin modules

justinbmeyer:
This implements the tilde extension `src/system-extension-tilde.js` which makes it easy for plugins to define places that can be replaced by `normalized` and `located` modules.  The idea is that a plugin can identify a regular expression (or an array of regular expressions) which will be used to replace moduleNames located in those regular expressions with a full address which has been found by running the module name through normalize and locate hook.

This applies replacements for less and css as outlined in issue #277


test/ext/css.js
```diff
@@ -70,6 +70,6 @@ function isProduction(){
 	return (loader.isEnv && loader.isEnv("production")) ||
 		loader.env === "production";
 }
-
```
matthewp: didn't you take this option away?

src/system-extension-tilde.js
```diff
@@ -0,0 +1,119 @@
+var addTilde = function(loader){
```
matthewp: can we rename this to something else, `system-plugin-locate.js` perhaps?

matthewp:
I'm wonder if this should be a configDependency rather than in core. I'm not sure about adding 120LOC for such a specific use-case, seems like it would make a nice configDep.

matthewp:
What would be even nicer is if we had a proper extension system as described in https://github.com/stealjs/steal/issues/414 and the [StealJS roadmap](http://forums.donejs.com/t/stealjs-roadmap-discussion/131), that way this could be pushed to NPM. But unfortunately for now moving this into `ext/` and adding it as a configDep is probably the way to go.

matthewp:
This is currently merged into minor (along with progressive package.jsons) so if you need to do any more prerelease let me know.

pYr0x:
to be clear, this PR is replaced by https://github.com/stealjs/steal/issues/605 ?

matthewp:
That's right.

Issue #578, opened 2016-02-13
tones of questions why steal and steal-tools are not working


pYr0x:
i tried my best to explain the problems i have with steal and steal-tools
http://forums.donejs.com/t/many-problems-with-multipage-app-build/150

i know your focus is not to much on building multi page apps.
but i think there is a community out there that havent a node server for the backend.

matthewp:
@pYr0x Are all of your issues fixed now, can we close this?

pYr0x:
check this out today and give you a feedback

pYr0x:
i could solved many problems
http://forums.donejs.com/t/many-problems-with-multipage-app-build/150/2?u=pyr0x

steal-sass is still not working as expected!

PR #579, opened 2016-02-14
Script tag wins against global steal object

pYr0x:
with https://github.com/stealjs/steal-tools/pull/370 now css files should loaded correctly and baseUrl should be not overwritten
https://github.com/stealjs/steal/blob/master/src/config.js#L324

this fix some of the problems with
stealjs/steal#578

### DIFF
--- a/src/startup.js
+++ b/src/startup.js
@@ -43,13 +43,15 @@
 			};
 		}
 
+		// first set the config that is set with a steal object
+		if(config){
+			System.config(config);
+		}
+
 		// B: DO THINGS WITH OPTIONS
 		// CALCULATE CURRENT LOCATION OF THINGS ...
 		System.config(urlOptions);
 
-		if(config){
-			System.config(config);
-		}
 
 		setEnvsConfig.call(this.System);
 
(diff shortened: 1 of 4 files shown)


matthewp:
Hi, thanks for submitting. Given that #578 probably covers several different issues please describe specifically what this commit fixes. It looks like it is related to bundleSteal, what specifically does this fix?

src/startup.js
```diff
@@ -43,13 +43,15 @@
 			};
 		}
 
+		// first set the config that is set with a steal object
+		if(config){
```
matthewp: Why move this here?

pYr0x:
talked to @matthewp 
rewrite the PR. provide a test.
now we can write a global steal object like 
```
<script>
    steal = {
        foo: 'bar'
    };
</script>
```
and steal merge that into the System-loader. 

if a script tag attribute is provided, steal overwrite the global property from the steal object.
this is necessary because of this issue: https://github.com/stealjs/steal-tools/issues/377
we now set 
```
steal.baseURL = './';
```
on files where steal is bundled in. if the user would change the baseURL he can write into the script tag or config.

Issue #518, opened 2015-11-19
Contextual Module Extension


matthewp:
We use this in live-reload, but it would be nice to have a generic API for defining a contextual module.

A contextual module is one that knows about it's parent, so a different module is registered for each parent.  As an example:

### chat/main
```js
import foo from "foo";
```

If **foo** were a contextual module, then the registry would contain a module called **foo/chat/main** and another for each module that imports foo.

Here's a possible API, the example is a contextual module that does normalization.

```js
System.setContextual("normalize", function(parentName){
  return {
    __useDefault: true,
    "default": function(moduleName){
      return System.normalize(moduleName, parentName);
    }
  }
});
```

Then this module could be used like:

### chat/main
```js
var normalize = require("normalize");

// normalize correctly passes in this module's name
normalize("child").then(function(name) { ... });
```

justinbmeyer:
@matthewp I'm not super clear about the example.  The heading implies that _chat/main_ module does:

```
import foo from "foo";
```

And somehow this maps to `foo/chat/main`?   I don't understand how that would work.  I would assume it would map to something more like `chat/main/foo`.

matthewp:
> And somehow this maps to foo/chat/main?

That would be the fully normalized module name, yes.

>  I don't understand how that would work. I would assume it would map to something more like chat/main/foo.

I suppose either would work ok. How I laid it out in this issue is how live-reload currently does it.  The module name can be whatever we want it to be, as long as it doesn't conflict with a module name that might exist in the user's project (which I don't think either of these could).  I think I prefer `foo/chat/main` because I see this as like a submodule of `foo` not a submodule of `chat/main` (it's a context-aware version of foo).

justinbmeyer:
This originated out of `.component` right?

If I had a component module name like: `pmo/restaurants/list.component!`, I don't think the `viewModel` should be `viewModel/pmo/restaurants/list.component!`.  

We might be agreeing, but I have what is the "parent" in your example wrong.  It comes across that the parent is `chat/main` because it is importing `foo`.  Parents import their children.

matthewp:
No, this has nothing to do with .component. This originated from live-reload. I can see what you're saying about .component's child modules.  This is not like that, `foo` is not a child of `chat/main`. It's a dependency of `chat/main` that knows `chat/main` is importing it.

We needed it for this:

```js
var reload = require("live-reload");

// The callback is called whenever THIS module is reloaded
reload(function(){

});
```

[...]

PR #580, opened 2016-02-16
Adding System.setContextual

phillipskevin:
Closes https://github.com/stealjs/steal/issues/518.


matthewp:
I think you had a different version of some upstream projects than what is is minor, which is why there are all of the unrelated changes showing up. If you rm -rf node_modules bower_components and reinstall everything, do another build, should fix it I think/hope.

phillipskevin:
@matthewp Thanks. I didn't realize that's what was causing those changes. Should be good to go now.

main.js
```diff
@@ -392,6 +392,51 @@ if(typeof System !== "undefined") {
 	addTilde(System);
 }
 
+function addContextual(loader){
+  loader._contextualModules = {};
+
+  loader.setContextual = function(moduleName, definer){
+    // `definer` could be a function or could be a moduleName
+		if (typeof definer === 'string') {
+			this._contextualModules[moduleName] = this['import'](definer);
```
matthewp: Like we talked about yesterday, doing the import here means that the import happens when `setContextual` is called, even if no one is going to be using this module. We should just save the string name here and import it in `normalize` when it's needed.

phillipskevin: I was going to put that on the steal-clone PR. Do you want it on this branch instead?

matthewp: Yeah, this one should be, I wouldn't want to merge this without that fix.

phillipskevin:
@matthewp I've updated the PR to only import contextual modules when they are used if you want to take another look.

phillipskevin:
@matthewp any issue with merging this?

matthewp:
Nope, looks good.

Issue #198, opened 2014-05-28
Tests don't pass in Phantomjs


daffl:
The current tests seem to get stuck in PhantomJS. I set the [TravisCI build](https://travis-ci.org/bitovi/steal/builds/26235780) up to use Firefox but it might be worth looking into it.

matthewp:
Should we try with Phantom 2?

matthewp:
I don't think we care about running these tests in Phantom right now, so closing.

PR #589, opened 2016-02-22
fileUrls are relative #198

alexisabril:
This is for issue https://github.com/stealjs/steal-tools/issues/198


matthewp:
@alexisabril should send pr against https://github.com/stealjs/system-npm, the ones in `ext/` are copied from that.

PR #592, opened 2016-02-22
fileUrls are relative

alexisabril:
This is for steal-tools issue: https://github.com/stealjs/steal-tools/issues/198


matthewp:
Thanks.

PR #610, opened 2016-03-03
locate:// schemes

### LINKED ISSUE #605: Replace 'LOCATE' and '~' module path rewriting with 'locate://' and 'pkg://'
New syntax originating from https://github.com/stealjs/steal-less/issues/11 , though that issue is specific to LESS.

The plan is to replace the existing LOCATE and ~ path rewriting syntax with new path schemes that can accomplish more and have more familiar syntax. 

'locate://' will work like the existing syntax, importing resources from a module ('bootstrap/main.less') or relative to the file containing the import ('../shared/main.less', './style.less').

'pkg://' will also import resources from a module, but will do relative imports relative to the project root.

nlundquist:
Resolves #605. Replace tilde with locate:// this includes the recursive rewriting for less imports.


matthewp:
There are conflicts now, probably easy to resolve.

matthewp:
@nlundquist I'm merged your changes in steal-css and steal-less, We need to do a release of those and update this pr to point to that release.

PR #615, opened 2016-03-04
Documentation metadata bundle=false

pYr0x:
updates the doc http://stealjs.com/docs/load.metadata.html
now "bundle"=false is a possible metadata property
https://github.com/stealjs/steal-tools/pull/386


matthewp:
Cool thanks, this will be on 0.15

PR #620, opened 2016-03-08
some improvements on utils.moduleName.parse

pYr0x:
try to test if the tests pass on travis.
write tomorrow a issue about this PR

**have to test plugin parsing!**


matthewp:
Hey @pYr0x thanks but I don't think this is right, moduleName.parse should parse the module name, it shouldn't care about paths config (or any other config).  What bug are you trying to work around?

pYr0x:
@matthewp if found some issues that caused by this parsing.
the most problems if found is with npm and multi-main.
also if using npm and make a build and use the "main" script data in steal. the script is not found.

if i am at home to night, i will add a issue described the problem.

pYr0x:
after talking to @matthewp , we find out, that is not the right fix for the problem.

Issue #627, opened 2016-03-17
baseURL is set wrong if using NPM in production


pYr0x:
lets say you are in a working directory `/var/www/myspace/web/` have a package.json like that:

**package.json**
```
{
	"name": "multi-main",
	"main": "main",
	"version": "1.0.0",
	"system": {
		"directories": {
			"lib": "src"
		}
	}
}
```

and you made a build with:
```
multiBuild({
    config: __dirname + "/npm-multi-main/package.json!npm",
    main: "multi-main/app_a"
}, {
    quiet: true,
    minify: false
});
```

you **cant** do that to load the build:
````
<script
        src="../../node_modules/steal/steal.js"
        main="multi-main/app_a"
        env="window-production"></script>
```
you got an error: `GET http://localhost:8080/dist/bundles/multi-main/app_a.js 404 (Not Found)`
**notice:** the src of steal is 2 folder behind

this is because the baseUrl is set wrong.

you have to set the base URL and the config like:
```
<script
        src="../../node_modules/steal/steal.js"
        main="multi-main/app_a"
        base-url="./"
        config="./package.json!npm"
        env="window-production"></script>
```

but you won't that, because if you provide no config, steal will automaticly load `package.json`

the problem is in https://github.com/stealjs/steal/blob/master/src/config.js#L331


**btw:**
if you made a build with `	bundleSteal: true` in the bundled file at the first lines you fill find:
```
steal.baseURL = './';
steal.configMain = "package.json!npm";
```

pYr0x:
@matthewp 
`/Users/julian/PhpstormProjects/steal-tools/` is the root folder. 
`http://localhost:8080/test/npm-multi-main/app_a.html` is the url

steal is located at `/Users/julian/PhpstormProjects/steal-tools/node_modules/steal/steal.js`

the examples are:

* if we don't add the `base-url` and `config` tags -> **failed**
```
<script
        src="../../node_modules/steal/steal.js"
        main="multi-main/app_a"
        env="window-production"></script>
```
=> steal would try to load `GET http://localhost:8080/dist/bundles/multi-main/app_a.js 404 (Not Found)`

* if we add `base-url` and `config` tags -> **pass**
```
<script
        src="../../node_modules/steal/steal.js"
        main="multi-main/app_a"
        base-url="./"
        config="./package.json!npm"
        env="window-production"></script>
```
=> steal load `GET http://localhost:8080/test/npm-multi-main/dist/bundles/multi-main/app_a.js`

PR #628, opened 2016-03-17
Base url npm

pYr0x:
set the baseURL relative to package.json 
in this case, it is always `./`

this is only if no config-main is provided!

fix: https://github.com/stealjs/steal/issues/627

@matthewp if you agree with this PR, can you please run Grunt on your lokal pc, so that the changes are in steal.js and steal.production.js
i have no grunt installed.. thank you


matthewp:
I don't think we can accept this change. The reason why the baseURL is set to be next to the config is for the common structure of:

```
\ node_modules/steal/steal.js
\ package.json
\ page1
\ page2
  - \ demo.html
```

Let's say you were at `http://localhost/example/page2/demo.html`.  Your script tag would be:

```html
<script src="../node_modules/steal/steal.js"></script>
```

And your package.json is located at `http://localhost/example/package.json`.

This allows many subpages to share the same configuration.  Setting the baseURL to always be `./` prevents that and essentially assumes that all subpages have their own configuration.

pYr0x:
good objection.. close this..

Issue #630, opened 2016-03-20
baseURL is set wrong if using bundleSteal in production


pYr0x:
when using `bundleSteal=true` on the build. we set things like this:
```
steal.loadBundles = true;
steal.baseURL = './';
steal.configMain = "package.json!npm";
```
but the `baseURL` is overwritten by `stealURL` setter function.
https://github.com/stealjs/steal/blob/22c4ee67285d0d5a3c4a47d4673757c64b415fbf/src/config.js#L230
https://github.com/stealjs/steal/blob/22c4ee67285d0d5a3c4a47d4673757c64b415fbf/src/config.js#L331

it also sets unneccesary paths when we using steal production
https://github.com/stealjs/steal/blob/22c4ee67285d0d5a3c4a47d4673757c64b415fbf/src/config.js#L270

matthewp:
We should probably be checking if baseURL already exists, and only set it if not.

pYr0x:
@matthewp what do you thinking about
``` 
		if(typeof config === 'object'){
			// the url options are the source of truth
			config = extend(config, urlOptions);
		}else{
			config = urlOptions;
		}

		// set the config
		System.config(config);
```
instead of
https://github.com/stealjs/steal/blob/master/src/startup.js#L47-L53

matthewp:
Let me know if that works. It might be ok, scares me just a little bit. Would like to see if the tests all pass (and if the steal-tools tests all pass with it).

pYr0x:
steal 0.15.2 and steal-tools 0.15.5 now setting env to production if steal is bundled
https://github.com/stealjs/steal/pull/635
https://github.com/stealjs/steal-tools/pull/425

Issue #631, opened 2016-03-21
Setting configPath via steal var fails since 0.14


robertheessels:
In my project use this code to set the configPah:

```
var steal = {

"configPath":"http:\/\/rhwp.bullpool.com\/wp-content\/themes\/swifty-site-designer\/ssd\/js\/stealconfig.js",

"main":"..\/preview.js?swcv=ssd_\/*@echo RELEASE_TAG*\/",

"bundlesPath":"http:\/\/rhwp.bullpool.com\/wp-content\/themes\/swifty-site-designer\/ssd\/js"

};
```

This used to work (0.13.2 and older), but fails since 0.14.

matthewp:
Thanks, this is production, right? You said you get an error that it can't find stealconfig.js? What does your script tag look like?

robertheessels:
It is in dev. Not sure if it would work in production.

My stealconfig.js file is (and needs to be) located 1 level lower than steal expects it to be.

Script tag:
`
< script type="text/javascript" src="http://rhwp.bullpool.com/wp-content/themes/swifty-site-designer/ssd/js/libs/steal/steal.js?swcv=ssd_%2F%2A%40echo+RELEASE_TAG%2A%2F&amp;ver=1.0"></script>
`

Both this script tag and the var steal= code above is generated by WordPress code in PHP, that's why it looks a bit weird.

pYr0x:
@robertheessels can you post the URL steal would like to load ?

robertheessels:
steal - error loading @config. Error loading "stealconfig.js" at http://rhwp.bullpool.com/wp-content/themes/swifty-site-designer/ssd/js/libs/stealconfig.js
Not Found: http://rhwp.bullpool.com/wp-content/themes/swifty-site-designer/ssd/js/libs/stealconfig.js

So the /libs is here incorrect.

(steal itself is in http://rhwp.bullpool.com/wp-content/themes/swifty-site-designer/ssd/js/libs/steal)

pYr0x:
@matthewp i am a little bit confused.
lets say we have this folder structure
![image](https://cloud.githubusercontent.com/assets/1126711/14032327/49f9166a-f212-11e5-8120-0a0cb15e8ff4.png)
and **bar.html** looks like

robertheessels:
Sorry, I do not fully understand what you mean.

[...]

PR #636, opened 2016-03-22
steal object and url-options merged into one config object

pYr0x:
related to https://github.com/stealjs/steal/issues/630 

the config on startup is now merged 
```
config = extend(config, urlOptions);
```

url-options is the source of truth


matthewp:
Is there a scenario where this is needed?  I think the only time "config" exists is during the build or no?

pYr0x:
@matthewp this PR is not a priority fix.

with this fix we prevent call `System.config();` twice which can cause unknown side effects.

`config` also exits if we have a `steal` object

pYr0x:
this PR solved 2 problems:
1. now run `System.config(config);` only once. the config are now merged. before, we run the config setter twice which overwrites some configs. this solves  #631. the test covers that, that the `config-path` will be overwritten.
2. now with the merged config setting, the URL-options wins against the `steal` object.

@matthewp can you please make a steal.production again.

matthewp:
You are amazing @pYr0x, will merge this tomorrow, but I've made it a rule not to do releases on fridays so ping me on Monday and i'll do a patch then.

pYr0x:
@matthewp push...

matthewp:
What do you mean?

pYr0x:
this branch stealjs:urloptions-source-of-truth is never pushed into master?

matthewp:
Oh. Why are you not pointing your PRs to master?

pYr0x:
.. good question... i will now start to working on stealjs repo not on my fork... future PR will against master :)

matthewp:
cool, if you prefer to work on your fork you can still submit the PR against master

PR #640, opened 2016-03-25
Do not allow cloned loader to leak scope

matthewp:
When creating a new loader we need a completely new loader with no existing modules. Previously we were copying the existing modules over and deleting parent modules.

The problem with this method is that it means that modules share their scope with a cloned loader and the "real" loader. So anything that is modified within that scope within the cloned loader code... will modify the outside scope as well..

The solution is to no loader copy the existing loads. Rather since we are already retaining the source, that will be enough.


justinbmeyer:
@matthewp lets talk about this sometime next week.  I think we should be copying the modules.  I'm not sure what you mean by the "share they scope" though.  However, I think the same version of something like jQuery or `CanJS`  is important.  Otherwise, each test will have to re-import everything side-effectual, like a component again.

justinbmeyer:
@matthewp ( cc @akagomez ) a little more context.  This test doesn't work:

https://github.com/donejs/bitballs/blob/1e2526af515314a9af6541f470e881003a44ac46/public/components/tournament/details/details_test.js#L98

If the above `clone` isn't commented out: https://github.com/donejs/bitballs/blob/1e2526af515314a9af6541f470e881003a44ac46/public/components/tournament/details/details_test.js#L28

The reason is that by using `stache` loaded from the scope like this:

```
 var frag = stache('<tournament-details {tournament-id}="tournamentId" />')({
            tournamentId: 2
        }), 
        vm = $(frag.firstChild).viewModel(),
        countRows = function () {
            return $('.games-table tbody tr', frag).length;
        };
```

This is using a different version of `stache` with a different registry for `<tournament-details>`.  

--------

Actually, upon reading what the test does, the problem is that the test is simply not importing `tournament/details` statically.  

@akagomez simple fix.

justinbmeyer:
Despite figuring out in this case, I'm sorta concerned that this might be a bit confusing.  Could this be an option with clone?

matthewp:
Do you disagree with the test case? What do you do about modules that have state?

matthewp:
It could be an option, but what is the correct default?

akagomez:
> Actually, upon reading what the test does, the problem is that the test is simply not importing tournament/details statically.

I don't understand. `tournament/details` is imported here: https://github.com/donejs/bitballs/blob/1e2526af515314a9af6541f470e881003a44ac46/public/components/tournament/details/details_test.js#L2

Are you implying that I have to re-import it in the specific test that depends on it?

justinbmeyer:
I missed that. Ok, I'm back to being confused.

PR #642, opened 2016-03-25
No main warning if steal is bundled

### LINKED ISSUE #641: main attribute requested even with bundleSteal
for production I use ```bundleSteal:true``` but in the console I have a warning: ```Attribute 'main' is required in production environment. Please add it to the script tag.``` Im using ```steal 0.15.2``` and ```steal-tools 0.15.11```, if steal is bundled no need for ```main``` attribute with the script tag.

pYr0x:
if we are in production and dont added a `main` attribute to the script tag, we get a warning.
but the `main` is not needed if we have steal bundled with steal-tools `stealBundle = true`

this PR prevent to get a warning.
we can use this new feature to detect if steal is bundled in the file https://github.com/stealjs/steal-tools/pull/425

close #641

### DIFF
--- a/main.js
+++ b/main.js
@@ -1071,7 +1071,7 @@ function addEnv(loader){
 		// we only load things with force = true
 		if ( System.loadBundles ) {
 
-			if(!System.main && System.isEnv("production")) {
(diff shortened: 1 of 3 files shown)


matthewp:
Thanks!

PR #647, opened 2016-04-01
Make meta configuration be set deep

### LINKED ISSUE #646: meta config should be deep
If the format doesn't change, don't override existing config. For example:

```js
System.config({
  meta: {
    foo: {
      format: "global",
      deps: ["bar"]
    }
  }
});

System.config({
  meta: {
    foo: {
      format: "global"
    }
  }
});
```

The deps should remain.

matthewp:
This is so that if 2 config calls both set a module to be `global` and the first adds `deps` but the second does not, the deps remain.

This is necessary because of progressive loading of package.jsons, the order of config calls is not deterministic.

Closes #646


matthewp:
Not quite ready, there are still bugs.

PR #655, opened 2016-04-05
Source of truth

pYr0x:
this PR should resolve the problems with steal-tools. all tests in steal-tools should now pass with this PR

i tested steal-tools with this commit of my fork https://github.com/stealjs/steal-tools/commit/c18212b343b0554c1f55ffc2e720d5187d8ad635
travis pass: https://travis-ci.org/stealjs/steal-tools/builds/121008438

@matthewp but i dont know why there is a conflict with merging, i tried very thing but without success. can you please look into the problem...

the main changes are:
* remove `if(this.stealBundled && this.stealBundled === true) {` because `if(!this.loadBundles) {` does it better
* update the `loadBundles` setter


matthewp:
What issues is this supposed to fix again?

pYr0x:
this fix this [issue](https://github.com/stealjs/steal/issues/631) and your test yesterday to check steal-tools against master of steal https://github.com/stealjs/steal-tools/tree/fail

matthewp:
So what was going on there? This makes the url options the source of truth, so was the `steal = {}` options overriding the url options? If so, which ones?

pYr0x:
so very simple.
if a config is provided e.g. `steal = {}` we extend the config with the `urlOptions`
```
if(typeof config === 'object'){
			// the url options are the source of truth
			config = extend(config, urlOptions);
		}else{
			config = urlOptions;
		}
```

matthewp:
Would you mind adding a test with the breaking scenario? I still don't have it clearly in my head. That would help me review this better.

pYr0x:
there is a test included `test/stealconfig/dev.html`

matthewp:
But this PR doesn't include any changes to that test. It was passing before these changes, no?

pYr0x:
The test will not passing without the changes...

matthewp:
So this is a new test that's not in master? It's weird that github is not showing it as being a new test...

matthewp:
Yeah, weird, this commit has the test https://github.com/stealjs/steal/pull/655/commits/3a28d41eb271462306d7fa8226edd4c7ce123b78

src/config.js
```diff
@@ -228,17 +232,15 @@
 					stealPath = paths.join("/"),
 					platform = this.getPlatform() || (isWebWorker ? "worker" : "window");
 
-				// if steal is bundled we always are in production environment
-				if(this.stealBundled && this.stealBundled === true) {
-					this.config({ env: platform+"-production" });
-
-				}else{
+				// we dont want to set all the development paths if loadBundles is true
```
matthewp: Why not, what problem does it cause?

pYr0x:
@matthewp maybe we have to talk about that, it is hard to explain what i have done...
we can look into this in the next hangout

pYr0x:
close this for https://github.com/stealjs/steal/pull/757

PR #660, opened 2016-04-09
process.cwd() should return a string.

marshallswain:
If `process.cwd()` returns undefined, it breaks node's `path` module when it's used in the browser.  This polyfill for process.cwd now matches the one in the babel polyfill here: https://github.com/stealjs/steal/blob/master/ext/babel-polyfill.js#L55

You can test for this by importing the path module in a project then doing
```js
path.resolve(process.cwd(), '401.html');  // Should match the next line
path.resolve('/', '401.html');

path.resolve(undefined, '401.html');  // Throws an error
```

### DIFF
--- a/ext/npm-load.js
+++ b/ext/npm-load.js
@@ -104,7 +104,9 @@ var translateConfig = function(loader, packages, options){
 	var g = loader.global;
 	if(!g.process) {
 		g.process = {
-			cwd: function(){},
+			cwd: function(){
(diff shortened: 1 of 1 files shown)


matthewp:
Hey, thanks, this should go against the stealjs/system-npm project. I know, it's weird.

marshallswain:
@matthewp I moved it over.  Thanks.

PR #663, opened 2016-04-12
Upgrade to steal-less 0.2.2

matthewp:


matthewp:
@kylegifford you have an issue with this change?

PR #666, opened 2016-04-18
Upgrade to system-npm 0.6.5

matthewp:


ext/npm-utils.js
```diff
@@ -267,8 +267,6 @@ var utils = {
 			return  utils.path.removeJS(
 				(pkg.system && pkg.system.main)
 				|| (typeof pkg.browser === "string" && (utils.path.endsWithSlash(pkg.browser) ? pkg.browser+'index' : pkg.browser) )
-				|| (typeof pkg.jspm === "string" && pkg.jspm)
```
pYr0x: you want to remove jspm ?

matthewp: Yes, we removed the test. The reason is that the `jspm` config is intended for cases where you are using jspm like `jspm install foo`. 

The breaking case was bootstrap, which uses jspm config, but the file structure for their jspm package is different than their npm package.

If there are any cases were removing jspm config support would break a package let me know.

PR #676, opened 2016-04-29
Fixes #675. Reject promise when startup doesn't find stealconfig

### LINKED ISSUE #675: Logging error when the config doesn't load.
https://github.com/stealjs/steal/blob/master/src/startup.js#L95

Should we doing this? I would think not.

I think the bigger question is... if we can't load the config is this an error case that the programmer definitely needs to fix. If so, we shouldn't allow steal to continue on.

If not being able to load the stealconfig is not that big of deal, we may not want output anything to the console.

mjstahl:


matthewp:
Maybe those tests were depending on the broken behavior... I'll take a look though, probably a problem with the tests.

matthewp:
Sorry for the delay, going to look at this today.

matthewp:
Replacing this with #735, merging with latest fixed it.

Issue #680, opened 2016-05-04
Cannot load enzyme library


m-mujica:
I've set up an example app that uses https://github.com/airbnb/enzyme but I can't get it to work

1. Steal throws trying to load enzyme's [main](https://github.com/airbnb/enzyme/blob/603f07adb31032d9b917b49685276f39390cfb76/package.json#L5) which points to the build folder where there is an `index.js` (which should be loaded by default in a CJS environment). My workaround was to use a combination of [map/paths](https://github.com/m-mujica/enzyme-steal-example/blob/cd136a73053f1566b7e3f108e18545fc34ce59ee/package.json#L20) to make Steal load the right file. Should steal try to load `index.js` if `build/build` is missing?

2. Enzyme depends on cheerio which tries to load its `package.json` using `require('./package')`, I guess that the extension in this case is required by the new spec, right? fwiw, the cheerio folks [changed that](https://github.com/cheeriojs/cheerio/commit/7b59afbc7a6aa39376021593e011bf718350bcb4) but it's still not available in npm. The workaround in this case was to add a [map using the full module name](https://github.com/m-mujica/enzyme-steal-example/blob/cd136a73053f1566b7e3f108e18545fc34ce59ee/package.json#L16)

3. Once the previous errors were fixed, steal throws trying to load a couple of deps (`react/lib/ReactContext`, `react/lib/ExecutionEnvironment`....), the modules are indeed missing, looking at [enzyme's docs](https://github.com/airbnb/enzyme/blob/master/docs/guides/webpack.md) I found that for webpack they set these modules as `externals`; from the webpack docs:

  ```
  **externals** allows you to specify dependencies for your library 
  that are not resolved by webpack, but become dependencies of the output.
  This means they are imported from the environment during runtime.
  ```

  I think steal does not have support for this (browserify supports this as well). Should we add support   for `externals`?

4. Just for the sake of testing, I [mapped the missing modules of 3 to `@empty` ](https://github.com/m-mujica/enzyme-steal-example/blob/cd136a73053f1566b7e3f108e18545fc34ce59ee/package.json#L14-L15) and got a new round of errors :( 

<img width="1433" alt="screen shot 2016-05-03 at 21 32 14" src="https://cloud.githubusercontent.com/assets/724877/15013127/fb9272f0-11d3-11e6-8e64-5739d036437f.png">

[...]

PR #683, opened 2016-05-07
Upgrade to system-npm 0.6.9

matthewp:


matthewp:
This is a partial fix for #680

Issue #701, opened 2016-05-18
Make npmAlgorithm: flat the default


matthewp:

matthewp:
Assigning it to @pYr0x since he's already worked on this.

PR #720, opened 2016-05-31
Default npm algorithm = "flat"

### LINKED ISSUE #700: Add npmAlgorithm: nested
Since flat will become the default in 1.0.

pYr0x:
for steal 1.0 npm algorithm will be flat which means that npm 3 or higher is recommended.
close #700 #701


package.json
```diff
@@ -19,7 +19,7 @@
   "devDependencies": {
     "steal-css": "0.4.1",
     "steal-less": "0.2.2",
-    "system-npm": "0.6.16",
+    "system-npm": "stealjs/system-npm#major",
```
matthewp: I'll do a prerelease of this so we can use npm instead of git.

pYr0x: ok... i will wait for it ;)

matthewp:
Just released system-npm `1.0.0-pre.0`, update the npm dependency to that.

pYr0x:
looks good ... ready to merge.. feel free to merge it

Issue #704, opened 2016-05-18
Remove script type=module "polyfill"


matthewp:
Since the real thing is [landing in browsers](https://blogs.windows.com/msedgedev/2016/05/17/es6-modules-and-beyond/) we shouldn't mess with it. We're doing the wrong thing currently. We should do https://github.com/stealjs/steal/issues/699 instead.

pYr0x:
how can i going to start at this issue?

matthewp:
Yeah, for this one the polyfill part is in https://github.com/stealjs/es6-module-loader, just remove it (and tests if they exist).  I'll comment on #699 on what to do.

matthewp:
Note that for https://github.com/stealjs/es6-module-loader we use the "base" branch, not "master". This is for historical fork reasons.

matthewp:
Closed by #729, in major.

PR #724, opened 2016-06-01
Move inner steal script module functionality to a script type called …

### LINKED ISSUE #699: Move inner steal script module functionality to a script type
This is an example of a bug caused by us doing this: https://github.com/stealjs/steal/issues/698

Instead of:

```html
<script src="steal.js">
  import foo from "bar";
</script>
```

Which causes the above bug, we could instead do:

```html
<script src="steal.js"></script>

<script type="text/steal-module">
  import foo from "bar";
</script>
```

pYr0x:
add a systemjs extension that will load modules within a script tag. after the dom is ready and all `configDeferred` and `devDeferred` are resolved

the type has to be "text/steal-module"

for legacy code, code within steals script tag is still allowed.

- [x] remove `modole`-tag support in es6-module repo (https://github.com/stealjs/es6-module-loader/pull/3)
- [ ] add docs 

close #699 and #704 is in process


matthewp:
I'll document in another PR. Leaving #699 open for now.

Issue #606, opened 2016-03-01
require a .js file do not load the file correctly


pYr0x:
if you would require a .js file
```
var MySalute = require("./salute.js");
```

steal technically speaking stealjs/es6-module-loader does not load correctly the file.
because in the `path` object there is a 
```
this.paths = { '*': '*.js' };
```
https://github.com/pYr0x/es6-module-loader/blob/js-file-bug/src/system.js#L154

matthewp:
Is this true if using the npm plugin?

pYr0x:
what do you mean by "using the npm plugin"?

matthewp:
If you're config is `package.json!npm` you are using the npm plugin and it's extension is added. I believe it's extension already allows you to include .js, but I could be wrong.

pYr0x:
@matthewp with the usage of the NPM plugin the module file is loaded correctly by importing like
```
var MySalute = require("./salute.js");
```

but if i using not the NPM plugin like this:
```
<script src="node_modules/steal/steal.js" main="@empty" config-main="@empty">
        var MySalute = require("./salute.js");
        console.log(MySalute);
    </script>
```
steal throws an error 
`steal.js:2648 GET http://localhost:8080/salute.js.js 404 (Not Found)`

should we provide the same behavior with and without the NPM plugin?
if yes, how can i solve this problem? where do i have to look into?

matthewp:
We could add that support, it would be a normalize hook that looks for .js and removes it (I suppose). Honestly, given how much we are pushing people to using npm, I don't know if it's worth the effort.

pYr0x:
hello @matthewp 
i wrote a normalizer hook for this bug.
but i have one question.

first my hook:
```
/*
 SystemJS JS-File
 Load a .js file Node-like
 */
function addJS(loader) {
    if (loader._extensions) {
        loader._extensions.push(addJS);
    }
    var normalize = loader.normalize;
    loader.normalize = function(name, parentName, parentAddress, pluginNormalize){
        var extension = ext(name);
        if(extension === "js" && !this.hasOwnProperty('npmContext')) {
            name = name.substr(0, name.lastIndexOf("."));
        }
        return normalize.call(this, name, parentName, parentAddress, pluginNormalize);
    };

    return loader;
}
if (typeof System !== "undefined") {
    addJS(System);
}
```
very simple.

but the problem is, that `mapping` with `System.map` is not covered.
 ```
steal.config({
	map: {
		"hello.js": "./salute.js"
[...]

PR #730, opened 2016-06-07
import js-modules with js-extension

pYr0x:
close #607 and #606 

now it is possible to load modules like in node
```
var Hola = require("./salute.js");
```

this fix does not interrupt the npm normalization 
```
var Hello = require("salute.js");
```
this will load a npm-package called `salute.js`


pYr0x:
@matthewp tests pass in travis, but i dont know why not in appvevor...

matthewp:
I'll take a look.

Issue #254, opened 2014-09-30
CSP build that loads with script tags


asavoy:
> NOTE: This is a request to make the StealJS format able to work like AMD and use purely a &lt;script&gt; tag to load StealJS modules.

@justinbmeyer mentioned this somewhere as a possibility. It would be great as a workaround, even if not the default behavior.

It would alleviate these issues:

 a) Browsers other than Chrome don't give usable error stacktraces/debug them nicely. This makes it _really hard_ to debug issues. (#243)

b) If you serve your JS from a different domain (e.g. with a CDN) you must ensure each file has the CORS Access-Control-Allow-Origin header correctly. Sometimes its preferable to load a third-party's script, e.g. Google Analytics, and they don't set this header.  (#251)

c) The meaning of document.write() changes. If a library makes use of it, it will overwrite the whole document, instead of inserting into it. Some libraries do this, and so loading with StealJS will break the page in strange ways.

**UPDATED:**
d) IE8/IE9 cross-domain XHR is strangely buggy, and some users can't load modules at all this way  - https://github.com/ModuleLoader/es6-module-loader/pull/224#issuecomment-64109163

matthewp:
@whitecolor that seems really useful, have you published this extension anywhere?  I was thinking about having a place that links to plugins.

whitecolor:
@matthewp no, I haven't publish publish it, you can do it yourself.

matthewp:
A CSP build is becoming more requested and I think there is some legitimate reasons to favor it.  Paired with https://github.com/stealjs/steal/issues/366 this would make a nice dev workflow for those who want to avoid eval.

whitecolor:
Yes, still waiting for this!

reasn:
I stumbled upon this issue while trying to create an application that runs both in electron and cordova. Electron brings its own `require()` while cordova relies on fully bundled and concatenated javascript.
Being able to load local files without chrome's XHR warnings allows to develop such applications in the browser without the need for watchers and development servers.

In short: This would be great ;)

matthewp:
Hey everyone, this has been a long time coming, I know. In steal-tools 1.4.0 we released the first version of optimized builds: https://github.com/stealjs/steal-tools/releases/tag/v1.4.0

These builds do not use XHR and should work fine in CSP environments. You can learn more [in this guide](https://stealjs.com/docs/StealJS.guides.optimized_builds.html). If for some reason there are CSP issues, I'll reopen this issue until we can resolve them.

PR #737, opened 2016-06-08
Use new Function for eval

### LINKED ISSUE #470: Unable to debug in Firefox
_can-connect_canjs_tests

I always have to clear my history after every change.  I'm not sure if there's anything StealJS is doing, or anything that we can do besides turn on cache-busting, but this is really annoying.

Serving files with http-server.

matthewp:
This updates Steal to use `new Function(source).call(context)` to evaluate modules. This method brings a better debugging experience for Firefox users.

Adds a new meta option `eval` that works like:

```
"system": {
  "meta": {
    "module": {
      "format": "global",
      "exports": "Foo",
      "eval": "script"
    }
  }
}
```

Documentation is included here. Closes #470


pYr0x:
@matthewp this is only in Dev mode? In production we don't need to evaluate scripts source?

matthewp:
We do, yeah.  https://github.com/stealjs/steal/issues/254 is for that problem.

PR #743, opened 2016-06-09
add test for module-detection #649

### LINKED ISSUE #649: `define` function indside the module make steal break
Say we have a module that inside itself defines function
```js
 return function define(definitions) {
    const createHref = makeCreateHref(namespace, _createHref)
    const match$ = history$.map(location => {
```
This is example:
https://github.com/TylorS/cyclic-router/blob/master/src/definitionResolver.js#L18

Than this module will be interprite by steal incorrectly and will not be loaded.

pYr0x:
change order of steal/systemjs module-detection

@matthewp i changed the order of https://github.com/stealjs/steal/compare/major...649-module-detection#diff-3a8c5737bc134c6682d08b218975af12L4812 will load the modules correctly.
unfortunately i cant create a new build of systemjs.. dont know what to do with MAKEFILE
can you change the order in systemjs and make a new release and update this PR?

close #649


matthewp:
Yep, will try to look at this tomorrow.

matthewp:
@pYr0x This breaks some tests in our systemjs fork. This is because a common pattern in AMD is:

```js
define(function(require){
  var foo = require("bar");
});
```

So the reordering will cause this to be detected as CommonJS.

I think to fix that bug we need to improve the regular expression that detects AMD modules which is here: https://github.com/stealjs/systemjs/blob/master/lib/extension-amd.js#L15

That's a hard regexp to swallow!  But it would be the way to go.

PR #749, opened 2016-06-09
Bring Node builtins into core

matthewp:
This brings the Node builtins into core Steal for 1.0. This means the user will no longer need to install steal-builtins and can do things like `require("http");` in their code (or more likely, dependency code).


matthewp:
This is a multi-part PR, more is coming.

daffl:
I approve very much.

Issue #754, opened 2016-06-10
npm plugin should create npm moduleNames always


matthewp:
Currently we only do so if the user is using directories.lib: https://github.com/stealjs/system-npm/blob/38f2403300ffd323e90e6beecf3447beaff06301/npm-load.js#L86

matthewp:
This is in the prerelease.

PR #755, opened 2016-06-11
npm plugin always forces System.main to have package name

matthewp:
This is to fix inconsistencies and potential double-loading where package name is used some times, and some times not.


matthewp:
Closes #754

PR #757, opened 2016-06-12
steal startup and config

### LINKED ISSUE #631: Setting configPath via steal var fails since 0.14

### LINKED ISSUE #644: Exception thrown in Chrome when Sidekick extension is installed
On Chrome, when users enable the "Sidekick by Hubspot" extension, Steal throws an uncaught exception.  This causes our app to fail loading.  Only happens when Steal is bundled with the application.
http://www.getsidekick.com/

```
steal.production.js:6 
Potentially unhandled rejection [37] Error loading ...

steal.production.js:6 
Potentially unhandled rejection [41] Error loading ...

```
Steal: 0.13.0
Steal-Tools: 0.13.2
_but also reproducible in latest versions_

You can easily reproduce by creating the Hello World app instructed here: http://stealjs.com/docs/StealJS.quick-start.html  The only one change you need to make is to set the bundleSteal option to true.

The problem appears to be due to the extension injecting a script tag.  It's breaking this part of the code (sorry that it's the minified version).

`        var O = function() {
            var a = {}
              , b = document.getElementsByTagName("script")
              , e = b[b.length - 1];
            if (e) {
                a.stealURL = e.src,
                d(e.attributes, function(b) {
                    var d = c(0 === b.nodeName.indexOf("data-") ? b.nodeName.replace("data-", "") : b.nodeName);
                    a[d] = "" === b.value ? !0 : b.value
                });
                var f = e.innerHTML.substr(1);
                /\S/.test(f) && (a.mainSource = f)
            }
            return a
        }
`

In the 'b' array, the "steal script tag" is not at the end, the extensions injected script tag is.

pYr0x:
* now the startup method is promise based. this solves the issue #644 where a script is added async to the  DOM and steal does not get URL configs.

* as second i consolidate the 2 times config setter here https://github.com/stealjs/steal/blob/master/src/startup.js#L46-L53 . we now have a config setter order. this close https://github.com/stealjs/steal/issues/631

* script tags config always wins !

* created a PR for steal-tools with this new steal https://github.com/stealjs/steal-tools/pull/476

* remove some deprecated and not documented config stuff like
`root` and `startId`.
`baseURL` and `baseuRL` gets converted into  `baseUrl` as seen in https://github.com/stealjs/steal/pull/757/files#diff-13f02842d60e5958db69f27f0d8eee8aR23

#### config order, from top to down
1. query config like `steal.js?main=foo&env=production`
2. steal object like `var steal = { main: "foo", env: "production"}`
3. script tag config like `<script src="steal.js" main="foo" env="production"></script>`

__so script tags config wins ever!__

this fix also close #644, #631 and the outdated #655


src/startup.js
```diff
@@ -1,132 +1,149 @@
...
 
-		} else {
-			configDeferred = System["import"](System.configMain);
+			// set the config
+			System.config(config);
 
-			devDeferred = configDeferred.then(function(){
-				setEnvsConfig.call(System);
-				setupLiveReload.call(System);
+			setEnvsConfig.call(this.System);
```
matthewp: `this` is no longer going to be `steal` since it's inside of a Promise then function.  Can put `var steal = this` before the `getUrlOptions()` call and refer to that.

src/startup.js
```diff
@@ -1,132 +1,149 @@
...
-				// This means we call it twice, but that's ok
-				if(config) {
-					System.config(config);
-				}
+			// Read the env now because we can't overwrite everything yet
 
-				return System["import"]("@dev");
-			});
+			// immediate steals we do
+			var steals = [];
```
matthewp: What is this for? I don't see it being used by anything.

src/startup.js
```diff
@@ -1,132 +1,149 @@
...
-				// if there's a main, get it, otherwise, we are just loading
-				// the config.
-				if(!System.main || System.env === "build") {
-					return configDeferred;
-				}
-				var main = System.main;
-				if(typeof main === "string") {
-					main = [main];
-				}
-				return Promise.all( map(main,function(main){
```
matthewp: This was to support multi-mains (where main is an array), did you remove support for this on purpose?

src/startup.js
pYr0x: you are right... will change that

src/startup.js
pYr0x: i didnt removed it.. it is only on a other position
https://github.com/stealjs/steal/blob/startup-and-config/src/startup.js#L120-L132

matthewp:
@pYr0x Let me know when you think this is ready again and i'll give it another review.

pYr0x:
i have some trouble in steal tools. if does not pass all tests.
we need to talk after i investigate

pYr0x:
@matthewp i did some refactoring with the config-setter. now there is a array of setter so that we have a order we go through each setter.
in steal all tests pass but in steal-tools i get some errors.

i made a branch that points to this steal PR.
https://github.com/stealjs/steal-tools/tree/new-steal-startup

unfortunately i didnt know what i can do to make the tests in steal-tools passing. it would be great if you can look into this.

i you want we can talk tomorrow.

matthewp:
Sorry for the delay, will re-review this today.

Issue #747, opened 2016-06-09
Document how to load scripts from CDN


phillipskevin:
For loading scripts from a CDN only in production, use envs config:

```
"system": {
  "envs": {
    "window-production": {
      "paths": {
        "grecaptcha": "https://www.google.com/recaptcha/api.js"
      }
    }
  }
},
"paths": {
   "grecaptcha": "grecaptcha.js"
}
```

For loading scripts from a CDN all the time, add the script with a `<script>` tag, and set it as instantiated:

```
<script>
  steal = {
    instantiated: {
      grecaptcha: grecaptcha: { 'default': window.grecaptcha, __useDefault: true }
    }
  };
</script>
```

matthewp:
Thanks!  Good contribution for anyone that needs one...

phillipskevin:
See https://gist.github.com/matthewp/083d9fecf1ebfc98d061218ea85b905f for a working example.

phillipskevin:
For SSR, you also need env configs for `server-development` and `server-production`.

PR #762, opened 2016-06-16
Adding docs showing how to load scripts from a CDN

phillipskevin:
Closes https://github.com/stealjs/steal/issues/747.

### DIFF
--- a/docs/pages/loading-from-cdn.md
+++ b/docs/pages/loading-from-cdn.md
@@ -1,0 +1,86 @@
+@page StealJS.loading-from-cdn Loading Scripts from a CDN
+@parent StealJS.guides
+
+@body
+
+Steal can be used to load script files from a URL, such as a Content Delivery Network (CDN).
+
+
+## Loading from a CDN in production
+
+A common use case is to load a script from a CDN in production-mode only.
+This example will show you how to load jQuery from NPM in development mode, but use jQuery's CDN in production.
+
+There are two changes that need to be made for this to work:
+
+* Set the System config to use the CDN path for production
+* Modify the build script to ignore jQuery when creating production bundles
+
+### System configuration
+
+Here is an example showing how to set the System config to load jQuery from production.
+
+```
+  "system": {
+    "envs": {
+      "window-production": {
+        "paths": {
+          "jquery": "https://code.jquery.com/jquery-2.2.4.min.js"
+        }
+      }
+    }
+  }
+```
+
+### Build script
+
+Update the build script to ignore jQuery.
+This example shows how to do this with a simple build.js file.
+
+```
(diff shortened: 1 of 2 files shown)


phillipskevin:
@matthewp let me know if any of the wording could be more correct. Also, I think you mentioned needing to set `server-development` and `server-production` for SSR to work, but I wasn't able to reproduce any issue with SSR without these.

docs/pages/loading-from-cdn.md
```diff
@@ -0,0 +1,86 @@
...
+* Configure Steal to use the previously loaded file
+
+### Load Using a <script> Tag and Configure Steal to use Loaded Version
+
+```
+    <script src="https://code.jquery.com/jquery-2.2.4.min.js"></script>
+    <script>
+      steal = {
+        instantiated: {
+          jquery: { 'default': window.jQuery, __useDefault: true }
```
matthewp: Yuck!  I'd like to figure out how we can fix this so the user doesn't have to know about the nasty internals like this.

docs/pages/loading-from-cdn.md
```diff
@@ -0,0 +1,86 @@
...
+        }
+      };
+    </script>
+    <script src="./node_modules/steal/steal.js"></script>
+```
+
+### Prevent Steal from Looking for jQuery During Build
+
+In order for the production build to work, map jQuery to the `@empty` so Steal can find it during the build:
```
matthewp: I would rephrase this slightly: `so Steal knows it doesn't need to try and find it.`

matthewp:
How about also linking to these docs from the [Moving to Production](https://raw.githubusercontent.com/stealjs/steal/master/docs/pages/production.md) docs.

docs/pages/loading-from-cdn.md
phillipskevin: I agree. Should we create a separate issue to make something like `instantiated: { jquery: window.jQuery }` work?

phillipskevin:
I added a link to the Moving to Production docs.

docs/pages/loading-from-cdn.md
matthewp: Yeah, there might already be one, ignore this comment as having anything to do with the doc changes.

pYr0x: would this 
```jquery: { 'default': window.jQuery, __useDefault: true }```
let me use jquery in a module regular?

e.g.:
__foobar.js__
```
var $ = require('jquery');
$('body').append('hello');
```


also, do i have to set jquery as a dependency in package.json?

matthewp: No, you don't need it as a dep.  `instantiated` puts it in the module registry.

PR #777, opened 2016-07-05
Allow cookies to be sent over CORS requests

joshAppDev:
I'm using steal to fetch libraries from another site. Some of them require authentication with cookies. This small change will enable that. (By default CORS requests will have cookies and auth headers stripped out by the browser.)

### DIFF
--- a/steal.js
+++ b/steal.js
@@ -2639,6 +2639,7 @@ function logloads(loads) {
         }
       };
       xhr.open("GET", url, true);
+      xhr.withCredentials = true;
 
       if (doTimeout)
         setTimeout(function() {


pYr0x:
@joshAppDev thx for contribute. But unfortunately this is not the correct why of doing it.
1. https://github.com/stealjs/es6-module-loader will be the right place
2. it is not supported by IE 10
3. we should make the xhr configurable, like jquery did http://api.jquery.com/jquery.ajax/ see user config here: http://stealjs.com/docs/steal.html#section_Configure
4. add a test, that test does not need to make a CORS request, but should check if the config is set right and the xhr object has `xhr.withCredentials = true;` maybe

PR #789, opened 2016-07-18
fixing issue with displaying CDN docs

phillipskevin:


docs/pages/loading-from-cdn.md
```diff
@@ -55,7 +55,7 @@ If you want to load a script from a CDN in all environments:
 * Load the file using a script tag
 * Configure Steal to use the previously loaded file
 
-### Load Using a <script> Tag and Configure Steal to use Loaded Version
+### Load Using a script Tag and Configure Steal to use Loaded Version
```
matthewp: You should be able to replace it with `&lt;script%gt;` instead.

phillipskevin: Thanks.

PR #811, opened 2016-07-27
Upgrade live-reload ext to 1.5.3

ccummings:
Enables live-reload in apps using https:// (wss://).


matthewp:
This is in `major`, it's backwards incompatible with 0.16.3.  If you need it in 0.16.x I could probably do a new live-reload release that only has the wss changes though.  Let me know.

Issue #772, opened 2016-06-30
Babel throws error for using deprecated option


akagomez:
Using this list of dependencies:

```
"steal": "git://github.com/stealjs/steal#v1.0.0-pre.7",
"steal-jsx": "git://github.com/stealjs/steal-react-jsx#v0.0.3",
"steal-tools": "git://github.com/stealjs/steal-tools#v1.0.0-pre.1"
```

I get this error in my Steal + React app: 

![](https://i.imgur.com/FjkAJWW.png)

Unless I add `delete options.blacklist` here: https://github.com/stealjs/steal/blob/master/steal.js#L2523

matthewp:
Probably because of this: https://github.com/stealjs/steal-react-jsx/blob/master/package.json#L18-L20  steal-jsx isn't currently compatible with 1.x

pYr0x:
i think we should remove `optional`, `whitelist` and `blacklist` in `options` if we are using babel >= 6?
so we dont get an error by react.
also remove the `babelOptions` in steal-react-jsx and update the readme. Users should use version `0.0.3` for steal < `1.0.0`

pYr0x:
fixed in steal 1.0

PR #826, opened 2016-08-13
babelOptions docs

pYr0x:
document which are the default babel-options set by steal

this PR is the last part for closing https://github.com/stealjs/steal/issues/772


matthewp:
Awesome! Thanks, these are great docs

PR #841, opened 2016-08-25
remove querymain

### LINKED ISSUE #840: Remove the main being set by a querystring in the steal src attribute
Instead to set main use `main="app/main"`. Some tooling uses query strings for various reasons, such as cache busting.

pYr0x:
the ability to load a module like `steal.js?moduleName,env` is gone.
use instead `steal.js?main=moduleName&env=development`

close #840

@matthewp pls review and merge


test/webworkers/main.js
```diff
@@ -1,4 +1,4 @@
-
+debugger;
```
matthewp: can you remove this?

matthewp:
Really we should be telling people to use the main attributes; and to use the attributes for configuration in general. Web Workers is the exception here.

test/webworkers/main.js
pYr0x: sorry.

i had review the code 2 times but didnt find that mistake :)

PR #847, opened 2016-09-23
Shim other frequently used console methods

daffl:


matthewp:
Need this in right away?

PR #848, opened 2016-09-23
Shim more console methods

matthewp:
This brings back the methods that were in https://github.com/stealjs/steal/pull/847


pYr0x:
Ups, what was with the first PR, there was a huge code delete...

matthewp:
It was run against some old versions of deps.  We should move running `grunt` only when doing releases.

PR #906, opened 2016-11-07
Update .travis.yml

frank-dspeed:
We Should realy think about testing on current ? 7.0.0 as 9 is prepared at present?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: node_js
 node_js: 
-  - 0.10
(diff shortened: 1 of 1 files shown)


matthewp:
Yeah, we can add current, not sure what travisci expects for that.

matthewp:
This is a breaking change, so can't go in master. I'm working on the `major` branch which will be 1.0 so this kind of change can go there. We want to support:

```
0.12
4.x
latest
```

PR #908, opened 2016-11-07
moved ecosystem from steal to stealjs

ilyavf:


matthewp:
Can you submit this against the `major` branch? `master` is for the 0.16.x version.

PR #936, opened 2016-12-01
npm plugin no longer crawls the first-level deps

matthewp:
This change makes it so that, when using the npm plugin, your first
level dependencies are no longer crawled. This means any plugins that
you use will need to be explicitly added in the steal.plugins
configuration:

```json
"dependencies": {
  "steal-less": "1.0.0"
},
"steal": {
  "plugins": ["steal-less"]
}
```

Using this config, steal will know to fetch steal-less' package.json,
      which will cause the configuration to be applied.

### DIFF
--- a/ext/npm-convert.js
+++ b/ext/npm-convert.js
@@ -34,7 +34,7 @@ exports.forPackage = convertForPackage;
 
 // Translate helpers ===============
 // Given all the package.json data, these helpers help convert it to a source.
-function convertSteal(context, pkg, steal, root, ignoreWaiting) {
+function convertSteal(context, pkg, steal, root, ignoreWaiting, resavePackageInfo) {
 	if(!steal) {
 		return steal;
 	}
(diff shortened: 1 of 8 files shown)


matthewp:
Blocked by https://github.com/stealjs/steal-npm/issues/203 for now.

matthewp:
stealjs/steal-npm#203 was fixed, running tests against steal-tools.

pYr0x:
What's the difference between `configDependency` And the `plugin` config?
Is steal-conditional a plugin or a config dependency?

matthewp:
For now `plugins` are things like steal-less that set up `ext` configuration. In the figure we might extend it. `configDependencies` are things that run before the NPM plugin is attached; they cannot import NPM packages and use them, so it's mostly only for low-level things like live-reload, and yes, steal-conditional.

pYr0x:
Do we need steal-stache also put into the plugin config?

matthewp:
Yep. Anything where you're not loading JavaScript.

pYr0x:
So we have to update https://canjs.com/doc/guides/setup.html#StealJSandnpm for steal 1.0.
@chasenlehara

matthewp:
Good point, I filed an issue on canjs:https://github.com/canjs/canjs/issues/2849

Issue #972, opened 2016-12-22
steal.startup no longer resolves to the main module


phillipskevin:
In Steal 1.0, `steal.startup()` resolves with no value here: [startup.js#L198](https://github.com/stealjs/steal/blob/master/src/startup.js#L198) (which calls [system-extension-script-module.js#L23-L31](https://github.com/stealjs/steal/blob/4ebf89f8d85b0542fddf7143a0ec7e145e06c521/src/system-extension-script-module.js#L23-L31)). In 0.16, it would resolve to the main module [v0.16.43/main.js#L1186](https://github.com/stealjs/steal/blob/v0.16.43/main.js#L1186)).

`done-ssr` uses the main module returned by `steal.startup`: https://github.com/donejs/done-ssr/blob/major/lib/startup.js#L70-L74. 

Since this changed, done-ssr is breaking when using steal 1.0:

```
Potentially unhandled rejection [2] TypeError: Cannot read property 'importPromise' of undefined
    at getMainModule (/home/travis/build/donejs/done-ssr/lib/startup.js:73:13)
    at /home/travis/build/donejs/done-ssr/lib/startup.js:40:10
    at tryCatchReject (/home/travis/build/donejs/done-ssr/node_modules/steal-es6-module-loader/dist/es6-module-loader.src.js:1183:30)
```

Here is an example failing test run: https://travis-ci.org/donejs/done-ssr/jobs/186219877.

matthewp:
I think the main's value gets returned [here](https://github.com/stealjs/steal/blob/e8f07292abd588ea21345f275dcc0e116d9dd5c4/src/startup.js#L191), so should be able to just make sure that's the value that's returned ultimately.

phillipskevin:
Yeah, it is. I just need to make sure it gets passed through. Thanks for confirming.

PR #973, opened 2016-12-23
ensuring that `steal.startup()` resolves with the main module

phillipskevin:
Closes https://github.com/stealjs/steal/issues/972.


matthewp:
If you need to do a release you'll have to increment the version in package.json and bower.json manually. And then run grunt which sets the version in steal.production.js. Those are the only 3 files that should be modified. If anything else is staged there was a mistake

PR #977, opened 2016-12-28
remove the system steal format standalone

### LINKED ISSUE #974: why system-format-steal.js exits in root folder?
why does this file exits in the root folder
https://github.com/stealjs/steal/blob/master/system-format-steal.js

should i not be unter '/src' ?
https://github.com/stealjs/steal/blob/master/Gruntfile.js#L36
https://github.com/stealjs/steal/blob/master/Gruntfile.js#L48

pYr0x:
remove the standalone system steal format extension
close #974


matthewp:
Weird that there's so many formatting differences.

matthewp:
What actually changed here, hard to tell with the diff.

pYr0x:
sorry... i changed my tab indent to space indent

matthewp:
Can you run `grunt` which will generate steal.js and steal.production.js. That would expose any problems that might cause tests to fail.

pYr0x:
does travis not run grunt and will catch any problems?

i do locally run grunt build and grunt test:testee
i got no errors but warning for example
`(node:1060) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: No parent suite found for test {"file":"http://localhost:3996/test/contextual/test.html","id":"761d128ff247b15cb0b0be2a0f1f0e81","title":"okay"}`

matthewp:
Ah, ok, you're right, Travis will run it. Maybe we should start only doing it on releases then.

pYr0x:
but we need grunt because in some tests we copy the new created steal into a test folder

matthewp:
@pYr0x looks like there are some conflicts after I merged your other PR. Can you update this one?

Issue #975, opened 2016-12-28
simplify and reorder steal creation


pYr0x:
today we concat some files into `system-format-steal.js`
https://github.com/stealjs/steal/blob/master/Gruntfile.js#L41-L48

my suggestion is, get rid of system-format-steal.js creation and remove 
- src/system-format-start.js
- src/system-format-end.js

and rename `system-format-steal.js` to `system-extension-steal.js`

matthewp:
Can we combine this issue with #976 ? I think it's really about the same thing.

matthewp:
I believe the reason for the weird ordering is that system-format-start -> system-format-end is all within a common closure like:

```js
// system-format-start
(function(){

// some stuff

// system-format-end
})();
```

It's definitely weird but I think the closure is needed? Anyways, open to a better way for sure.

pYr0x:
i dont know why a IIFE is needed for the steal-extension. because we have other system extensions that work the same way.

i would prefer a refactor the whole `addExtension` stuff.
we have a lot of this code in steal
```
if (typeof System !== "undefined") {
	_SYSTEM_addJSON(System);
}
```
maybe it would be better to create a function called `addStealExtension`
that function can handle all the common stuff like
- is System not undefined
- add the extension to the `loader._extensions`

matthewp:
Sounds good to me:

```js
addStealExtension(function(){
  // Whatever goes here, this is the function that is pushed into loader._extensions
});
```

Also would be good to expose to 3rd part extensions like live-reload:

```js
var steal = require("@steal");

steal.addExtension(function(){
  // extension goes here.
});
```

PR #978, opened 2016-12-28
Refactor extensions

pYr0x:
PR for #975

comments are welcome.


matthewp:
Will review tomorrow, thanks!

pYr0x:
we should think about system-trace. it is a system extension like all other. we should put system-trace into core.

pYr0x:
Don't know why Travis failed, the error doesn't  say anything to me.

matthewp:
@pYr0x agree on system-trace coming into core.

pYr0x:
i can create a new issue and a PR for that

matthewp:
Ok, as far as the CI breaking, looks like 1 test is failing in a few browsers: https://saucelabs.com/beta/builds/2bc0082fe93d454c9ba6b97c1d6fdf46

pYr0x:
seems that the service worker test will fail

src/end.js
```diff
@@ -21,7 +21,6 @@
 				}
 			});
 		global.steal.clone = cloneSteal;
-		global.steal.addSteal = addSteal;
```
matthewp: There's another addSteal in this file, is it not needed any more either?

pYr0x: i searched for `.addSteal` in the repo and didnt find another reference

matthewp:
Would you mind pointing steal-tools to this branch and letting CI run on it, just to verify it doesn't mess with the build in any way?

pYr0x:
@matthewp seems that steal-tools will work with this PR
https://github.com/stealjs/steal/tree/refactor-extensions
https://travis-ci.org/stealjs/steal-tools/builds/187590727

Issue #985, opened 2016-12-30
steal-clone dont work with exporting a object


pYr0x:
i saw this test in can-observation
```
QUnit.test('should throw if can-namespace.Observation is already defined', function() {
	stop();
	clone({
		'can-namespace': {
			Observation: {}
		}
	})
	.import('can-observation')
	.then(function() {
		ok(false, 'should throw');
		start();
	})
	.catch(function(err) {
		ok(err && err.indexOf('can-observation') >= 0, 'should throw an error about can-observation');
		start();
	});
});
```

but this test fail.

the problem is, that the cloned `can-namespace` isnt giving a plain object back.
steal-clone is given a `Module` with a method called `Observation`

testing in can-Observation does work this way
```
if (namespace.Observation) {
	throw new Error("You can't have two versions of can-observation, check your dependencies");
} else {
	module.exports = namespace.Observation = Observation;
}
```

@phillipskevin @matthewp

phillipskevin:
This test https://github.com/canjs/can-types/blob/master/can-types_test.js#L39-L55 is working fine. Going to compare to see why the `can-observation` test is different.

phillipskevin:
Ok, both of these are behaving the same. Doing this:

```
	clone({
		'can-namespace': {
			Observation: {}
		}
	})
```

creates a Module with an `Observation` property that only has a getter. 

The issue with the `can-observation` test is that `can-observation` loads `can-event` which tries to set another property on `can-namespace` (`namespace.batch`). This cannot be done because we're preventing extensions, here: https://github.com/stealjs/steal/blob/master/steal.js#L2385-L2386.

This test needs to correctly override the default export of `can-namespace`, like:

```
	clone({
		'can-namespace': {
			default: {
			    Observation: {}
                        },
                        __useDefault: true
		}
	})
```

This is ugly, but it is the correct way right now to override a default export with `steal-clone` so that when you import the default (like `import namespace from 'can-namespace';`) you get the passed value instead of a Module.

Maybe we should do this automatically so that if you pass an object with a single property it will turn it in to the `default`. We might also need to do this if you pass a function like:

```
	clone({
		'can-namespace': function() { ... }
	})
```

Otherwise, we need to update the documentation since right now it does not show how to override the `default` export correctly.

pYr0x:
yep i like the idea of doing it automatically  with plain objects and functions

matthewp:
What about an ES module that only has one export like:

```js
export function Observation(){}
```

[...]

PR #992, opened 2017-01-06
Simplify using steal-clone with default exports

phillipskevin:
Previously to override a default export with steal-clone, you had to do:
```
clone({
  'jquerty': {
    default: {},
    __useDefault: true
  }
})
```

With this change, you do not need to pass `__useDefault` and steal-clone will set it for you.

Closes https://github.com/stealjs/steal/issues/985.

### DIFF
--- a/docs/module-steal-clone.md
+++ b/docs/module-steal-clone.md
@@ -42,7 +42,9 @@ clone({
       return 'moduleBOverride';
     }
   }
-}).import('moduleA', function(moduleA) {
+})
+.import('moduleA')
+.then(function(moduleA) {
   moduleA(); // moduleA moduleBOverride
 });
 ```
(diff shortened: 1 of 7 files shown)


pYr0x:
Thank you!
I think a short description in the docs about default export would be good

phillipskevin:
@pYr0x the first example in the docs shows overriding a default export. Do you think there is other information that would be useful?

matthewp:
Should this be a minor or patch release item?

phillipskevin:
I think it's just a patch release. It was documented to work this way, but didn't work correctly.

pYr0x:
@phillipskevin maybe a example that shows returning a plain object like can-namespace?

matthewp:
I'll release this on monday.

matthewp:
@phillipskevin this is backwards compatible I assume?

phillipskevin:
@matthewp yes, it is.

matthewp:
Ok, I have a patch release to do today and don't want 2 releases in the same day... so will do this one as 1.1.0 tomorrow.

phillipskevin:
👍 Thanks.

PR #1015, opened 2017-01-18
move steal cache-bust into core

pYr0x:
this PR moves steal's cache-bust extension into steal's core.

Todo:

- [ ] add docs


matthewp:
@pYr0x In what way is this a breaking change?

pYr0x:
the "old" cachebust extension you have to configure it like
```
...
"steal": {
  "cacheKey": "foobar",
  "cacheVersion": 1
}
```

pYr0x:
ideas:

- no special setter for cache version with script tag
- write out the version number on steal production
- while buidling, take the version number (in package.json) and add a hash of all bundles `1.0.0-CRAZYHASH`

matthewp:
To summarize our discussion today at the contributor meeting:

## Including a buildVersion in steal.production.js

If we were to include a buildVersion in steal.production.js that would cause steal.production.js itself to not come from the cache.  Let's say the browser loads these files:

- steal.production.js
- bundles/app.js
- bundles/bundle-a.js
- bundles/bundle-b.js

If something changes with `bundles/bundle-b.js` then that is the only file that should be loaded from network in production. If we add  a `buildVersion` flag to steal.production.js then that means that steal.production.js will also come from the network.

This is an undesirable consequence. 

At this point I think the Steal buildVersion feature is out-of-scope of cache busting. They are two separate things.  I've moved discussion for that feature over to this issue in steal-tools: https://github.com/stealjs/steal-tools/issues/587

I don't think it should block cache busting.

## Cache busting implementations

There are two ways I can think of to implement cache busting:

### client-side

The way this PR does it, by looking for a version number and including that as a query parameter.

### at build-time

It could also be part of the build. Let's say you have a build like:

- bundles/app.js
- bundles/bundle-a.js

And you did something like this:

```js
stealTools.build({}, {
  hashBundleNames: true
});
```

This would include a hash in the bundle names so that they instead were something like:

- bundles/app-HASH.js
- bundles/bundle-a-HASH.js

This would mean that any time the contents of the bundle changed, the hash would change, and the result would be that the cache would be busted. You wouldn't need the client-side cache busting at all.

matthewp:
@pYr0x is this ready to be reviewed again?

pYr0x:
@matthewp we first should solve the problem that the tests are not started (https://github.com/stealjs/steal-qunit/issues/22#issuecomment-275231682) with the mapping stuff here https://github.com/stealjs/steal/blob/0212dfdafe0b9f7da777c443840a3bcaafd177e0/src/cache-bust/test/test.html#L11-L21

matthewp:
Ok, i'll pull down the branch and look. The fact that it doesn't work with the mappings makes me think it's probably not a steal-qunit bug.

matthewp:
@pYr0x I fixed the configuration so that the tests will run properly. A couple are failing though.

pYr0x:
@matthewp ok, good to see....
my tests of cachebust are failing?

matthewp:
Yeah, I didn't investigate why.

pYr0x:
@matthewp ready to merge.
still have to write the docs

pYr0x:
going to merge this to work on https://github.com/stealjs/steal/issues/1027

matthewp:
@pYr0x plan is to release 1.2 with this and the globals config that @m-mujica is working on.

PR #1023, opened 2017-01-25
Moved the env-extension into core

pYr0x:
close https://github.com/stealjs/steal/issues/1019

PR #1032, opened 2017-01-31
Update steal-systemjs to 0.19.1

m-mujica:


main.js
```diff
@@ -1274,19 +1274,16 @@ addStealExtension(function (loader) {
...
-if(typeof System !== "undefined") {
-	addEnv(System);
-}
-
-function addEnv(loader){
-	// Add the extension to _extensions so that it can be cloned.
-	loader._extensions.push(addEnv);
+// Steal Env Extension
+// adds some special environment functions to the loader
+addStealExtension(function (loader) {
```
matthewp: I think you maybe had a stale node_modules, this change isn't related to the steal-systemjs upgrade.

m-mujica: @matthewp I did a clean install and this comes up when I run grunt, maybe something we forgot to update before?

pYr0x: @matthewp i think you forgot to make a release after merging https://github.com/stealjs/steal/pull/1023 the last steal release is 7 days ago. the PR is 6 days ago.

matthewp: That change didn't need a release, it was just code reorganization.

pYr0x: yep, but i didnt run the grunt task because i thought you will run a `npm publish`

matthewp: Ok, no problem. I usually always run the grunt task when sending a PR. It's not an ideal system but I can't think of a better way right now.

pYr0x: ok good to know. i will right now always run the build and then make a PR

PR #1033, opened 2017-01-31
refactor tests

### LINKED ISSUE #1027: some wrong with tests
@matthewp it looks something is going wrong with the tests

1. we dont run `testee` in travis anymore. in travis only the node-tests are triggered. https://github.com/stealjs/steal/blob/master/.travis.yml#L10
the `npm run test` was removed. 
@m-mujica and i added your extension-tests to the testee grunt task. so your tests will be actually not run.
2. sauce-labs only runs the main test of steal. https://github.com/stealjs/steal/blob/master/test/test-sauce-labs.js#L114

solutions: 
1. add testee tests to travis again
2. add the extension-tests to the main steal test like the `json` extension.

what do you prefer @matthewp ?

pYr0x:
refactor tests and enable them all on `test.js` so that saucelab will run all test in different browsers

this PR also contains the changes of the cache-bust extension, because i didnt run the `build` grunt task on that PR

lets see if all going right.

🎉  all tests pass

will close #1027


src/json/test.html
```diff
@@ -1,22 +1,41 @@
...
+</script>
+<script
+        src="../../../steal.js"
+        main="src/json/test/"
+        config-main="../../../package.json!npm"
+>
+</script>
+<script>
+	steal.done().then(function () {
+		QUnit.config.hidepassed = true;
```
matthewp: I like to see the tests as they run, personally. Some times I want to rerun tests even if they pass and this option prevents them from even showing up in the list.

pYr0x: it is ok, on saurcelab we run with ?hidepassed

PR #1053, opened 2017-02-09
Move steal-systemjs into core

### LINKED ISSUE #1052: Move steal-systemjs into core
Part of #746 

@pYr0x were you planning to work on this? Otherwise I'll do it 

@matthewp how many points would you add to this? I'd like to add it to bitcentive feb.

m-mujica:
Closes #1052


pYr0x:
@m-mujica can you refactor the tests like this
https://github.com/stealjs/steal/blob/e2086792f7a9dd1604e8a5bf811db6b6fc2f88a7/src/trace/trace_test.html

m-mujica:
@pYr0x I don't think we want to use steal to load the tests, since we only want to test that systemjs works fine. 

Steal will bring systemjs and other stuff.... makes sense?

matthewp:
Makes sense. 

Currently when running in Node steal-systemjs is required: https://github.com/stealjs/steal/blob/87b7bc7d72e2ed22e6c605cbe953327c3d6f5d4a/src/end.js#L2

I suppose now we'll want to include this code in the main.js build.

pYr0x:
@m-mujica but `"src/system/system.js",`  is a part of stealjs.
systemjs is always in combination with steals other stuff. i think i would be good to test if systemjs is working with steal. if we break some compatibility with systemjs these tests will not cover this.

m-mujica:
@pYr0x I see what you're saying but the rest of the tests (e.g https://github.com/stealjs/steal/blob/e5d26e2be1bfb91bc04bbf1ca7c4fe4986889059/test/test.js#L87) already make sure that steal "uses" system correctly.

Maybe bring it up tomorrow during the meeting for discussion but I think we should keep steal out of the system extension tests.

steal.js
```diff
@@ -5029,7 +5042,7 @@ var $__curScript, __eval;
 
   if(isWorker) {
     $__global.upgradeSystemLoader();
-  } else if ((isBrowser || isNW) && !isChromeExtension) {
```
matthewp: Looks like you don't have the latest es6ml in your local node_modules.

steal.js
```diff
@@ -4271,17 +4283,15 @@ function amd(loader) {
   var commentRegEx = /(\/\*([\s\S]*?)\*\/|([^:(?!\\)]|^)\/\/(.*)$)/mg;
   var stringRegEx = /("[^"\\\n\r]*(\\.[^"\\\n\r]*)*"|'[^'\\\n\r]*(\\.[^'\\\n\r]*)*')/g;
   var beforeRegEx = /(function|var|let|const|return|export|\"|\'|\(|\=)$/i
-  var cjsRequirePre = "(?:^\\uFEFF?|[^$_a-zA-Z\\xA0-\\uFFFF.\"\'])";
-  var cjsRequirePost = "\\s*\\(\\s*(\"[^\"\\\\]*(?:\\\\.[^\"\\\\]*)*\"|\'[^\'\\\\]*(?:\\\\.[^\'\\\\]*)*\')\\s*\\)";
```
matthewp: I made some changes to steal-systemjs recently, make sure you have the latest.

matthewp:
*Ideally* each extension would be in its own folder inside of `src/` as a sibling to all of the other extensions.  For example the AMD extension would be in `src/amd/`. But I realize this is a larger change. There's value in just getting steal-systemjs and steal-es6-module-loader into core so that it's easier to make changes to Steal.

Perfect is the enemy of good. So let's leave the tests as they are, and over time we'll move the individual extensions to their own folders and make everything work the way the steal extensions work.

One change I *would* like to see now is renaming this `system` folder. I want to expunge the word *system* from Steal :).  How about **base** instead? As these are the base extensions. When you move es6ml into core let's call that one **loader**. Fair?

m-mujica:
@matthewp sounds fair to me, I just pushed the changes requested.... Once I move the loader extension I'll work on moving each base extension to `src`.

screen shot 2017-02-11 at 11 14 51

matthewp:
This is going to be a big boon when doing bug fixes... @m-mujica can you update stealjs/systemjs with a deprecation notice?  Thanks!

PR #1063, opened 2017-02-17
Move es6-module-loader into core

### LINKED ISSUE #1056: Move es6-module-loader into core.
https://github.com/stealjs/steal/issues/746#issuecomment-278837024

m-mujica:
Closes #1056

This one is lots of fun! 😨 

- [x]  Migrate Grunt task
- [x] Get node tests running (removed the `expect.js` dependency 🥇) 
- [x] Delete `src/loader/tests/node/index.js` (deprecated)
- [x] Get browser tests passing 
- [ ] Deprecate es6-module-loader once this has been merged


matthewp:
These tests use a custom test runner iirc. Probably want to convert them to qunit.

m-mujica:
@matthewp the same tests are run both in node and in the browser (through karma) I tried to rewrite one of the tests to qunit but I can't get it to run in node.... QUnit's node api is terrible https://github.com/kof/node-qunit#command-line

I could either:

a) Rewrite tests to QUnit but only run them in the browser
b) Keep mocha and work on the browser setup with steal-mocha (get rid of karma)

thoughts?

matthewp:
I think it's ok to only run them in the browser. We probably do want to eventually have a way to run tests in both contexts (we do have a few Node tests) but we can tackle that in the future.

m-mujica:
iterm2

src/loader/lib/system.js
```diff
@@ -0,0 +1,297 @@
...
+      }).then(null, function(err){
+        reject(err);
+      });
+    }
+  }
+  else {
+    throw new TypeError('No environment fetch API available.');
+  }
+
+  class SystemLoader extends __global.LoaderPolyfill {
```
m-mujica: @matthewp I wonder if we should check in a transpiled version of this file, to get rid of the grunt task that uses `grunt-esnext` it seems to be not deprecated and it made me use an older version of grunt....

or we could maybe merge this thing and open a new issue. Tests are looking good.

m-mujica:
@matthewp 👍  or 👎?

matthewp:
👍

PR #1068, opened 2017-02-23
Get steal-npm into core

### LINKED ISSUE #1064: Move steal-npm into core

m-mujica:
Closes #1064


matthewp:
The intent is for `src/` to be for stuff that gets built into steal.js. Since we don't build the npm plugin, I think we should keep it in `ext/npm.js`, etc. and just edit those files directly. So there's just the matter of putting the tests somewhere. That's a hard one to decide... but I guess `test/npm` is probably the place.

m-mujica:
giphy

makes sense

m-mujica:
there is already a test/npm folder, test/npm-extension maybe?

matthewp:
We can get rid of the tets/npm tests, they were needed to test "integration" but now that it's in core that's needed. Seriously doubt they test anything that the steal-npm tests don't.

m-mujica:
Ok @matthewp I think this is in good shape for a second review! let me know what you think.

matthewp:
Do you know why loader.js has changes?

m-mujica:
@matthewp I think it's the Promise shim in when.js, I didn't lock down the version to  `3.7.5` as in the es6-module-loader repo, want me to do that?

m-mujica:
@matthewp yup, it's that, just locked down the version locally and saw the changes. Let me commit that

matthewp:
Lgtm

PR #1081, opened 2017-03-01
Load development bundles when flag is set

### LINKED ISSUE #1074: Automatically load dev bundle when the flag is set
```
<script src="node_modules/steal/steal.js" dev-bundle></script>
```

m-mujica:
Closes #1074

Need to update this PR to handle the new two development bundles:

- [x] Load `dev-bundle` AFTER `@config` when `deps-bundle` flag is set
- [x] Load `dev-bundle` BEFORE of  `@config` when `dev-bundle` flag is set

1)
```html
<script src="node_modules/steal/steal.js" deps-bundle></scripts>
```

2)

```html
<script src="node_modules/steal/steal.js" dev-bundle></scripts>
```


main.js
```diff
@@ -1474,7 +1474,12 @@ addStealExtension(function (loader) {
 				});
 
 			} else {
-				loader["import"](loader.configMain)
+				var devBundlePath = (loader.devBundleDest || "") + "dev-bundle";
```
m-mujica: @matthewp is there a helper already to join paths? so I don't have to add code to handle trailing slashes, and stuff like that?

matthewp: I think there's a steal.joinURIs

m-mujica:
deps + config

dev-bundle

m-mujica:
only deps 

deps-bundle

m-mujica:
@matthewp can you take a look when you have a chance, please?

docs/pages/development-bundles.md
```diff
@@ -0,0 +1,93 @@
...
+@parent StealJS.guides 2
+@outline 0
+
+@body
+
+<!-- hack! -->
+<style>.contents { display: none; }</style>
+
+In this guide we'll go through the steps required to set up `steal-tools` development bundles and
+speed up your application development process.
```
matthewp: I would add a note in here (like a blockquote) that says that we are assuming that the reader already has a project setup. Maybe refer them to the progressive loading guide if they don't already have a project to try this with.

docs/pages/development-bundles.md
```diff
@@ -0,0 +1,93 @@
+@page StealJS.guides.development_bundles Development Bundles
+@parent StealJS.guides 2
+@outline 0
+
+@body
+
+<!-- hack! -->
+<style>.contents { display: none; }</style>
```
matthewp: Is this to disable the TOC? Is that something we want to do?

test/dev_bundles/dev.html
```diff
@@ -0,0 +1,37 @@
...
+<head>
+	<meta charset="UTF-8">
+	<title>StealJS development bundles</title>
+</head>
+<body>
+	<script
+		dev-bundle
+		main="@empty"
+		src="../../steal.js"
+		dev-bundle-dest="test/dev_bundles/"
```
matthewp: I was thinking that `dev-bundle` and `dev-bundle-dest` should just be one option.

The way I would implement this is to have code [just like this](https://github.com/stealjs/steal/blob/master/src/config.js#L35):

```js
System.devBundle = "@empty";
```

Then setup a setter [on this object](https://github.com/stealjs/steal/blob/master/src/config.js#L180)

It would make it so that `dev-bundle` attribute becomes:

```js
System.devBundle = "dev-bundle";
```

And if the user types `dev-bundle="test/dev_bundles/dev-bundle"`

Then it becomes:

```js
System.devBundle = "test/dev_bundles/dev-bundle";
```

This way in startup.js you can just directly do `System.import(System.devBundle)` and not have to worry about checking if it exists and doing any sort of pathing stuff. It will be `@empty` by default, otherwise it goes through the normal config pipeline to set its moduleName.

matthewp:
1q3gh9j

PR #1089, opened 2017-03-06
Allow relative condition module names

m-mujica:
Need to update this PR once steal-conditional gets a patch release with the relative module name change.


matthewp:
What's the reason for testing steal-conditional in this repo?

m-mujica:
@matthewp should theses tests be added to steal-conditional itself? I remember we decided to write the steal-conditional tests in other repos but when I moved steal-npm, some steal-conditional tests ended up in steal itself.

PR #1095, opened 2017-03-09
Fixes incorrect baseURL in IE

### LINKED ISSUE #1094: baseURL is wrong if multiple <script> tags contain /steal/ in IE
loading http://localhost/canjs/node_modules/can-view-autorender/test/basics.html which contains these scripts:

```
<script src="../node_modules/steal/steal.js" main="@empty"></script>
<script src="../../../node_modules/steal/steal.js" main="@empty"></script>
```

even if the first script is a 404, `baseURL` will be set as though it was successful:

```
>> steal.System.baseURL 
"http://10.0.2.2/canjs/node_modules/can-view-autorender/" 
```

This happens in IE9, IE10, (probably IE11).

m-mujica:
This change mimics steal@0.16 behavior in browsers without [document.currentScript support](https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript); meaning, the last script tag on the page is assumed to be the one loading `steal.js`

The previous approach of attaching a `onload` handler to the the scripts fails in IE10 because `load` events are fired regardless the script succeeded to load or not. If multiple tags happen to be loading scripts with `src` matching `/steal/` the first to match would end up being used to set `baseURL`. 

Fixes #1094


phillipskevin:
:shipit:

m-mujica:
Test failed in both IE9 and IE10

win7_ie_10 0

with this branch 🎊 

ie9
ie10

PR #1100, opened 2017-03-13
Basics of Babel custom plugins through StealJS

### LINKED ISSUE #1040: Support customizing Babel / core JS transpiler
Ideally, this could be done via `plugins` so if someone loads ES8 code, it can work within someone's ES7 app, but my guess is that to start, we can make it one configuration.

Along with this, it would be good to allow transpiling of CJS modules. 

## Things to consider

- How would steal-tools be configured with the same transpiler?
- What if different plugins wanted their own extensions?

m-mujica:
Closes #1040

- [x] Support plugin options
- [x] Read default from plugin module if using ES6?


src/loader/lib/transpiler.js
```diff
@@ -2,121 +2,163 @@
...
+			delete options.optional;
+			delete options.whitelist;
+			delete options.blacklist;
+
+			options.presets = options.presets || [];
+			options.plugins = options.plugins || [];
+
+			// make sure presents and plugins needed for Steal to work
+			// correctly are set
+			options.presets.unshift("es2015-no-commonjs", "react", "stage-0");
```
m-mujica: this has to happen only once, otherwise we'll unshift the default for each module loaded.

src/loader/lib/transpiler.js
```diff
@@ -87,26 +91,83 @@
...
+			var name = typeof plugin === "string" ? plugin : plugin[0];
+			var parent = self.configMain || "package.json!npm";
+
+			if (!isPluginRegistered(babel, name)) {
+				var pluginPromise = self.normalize(name, parent)
+					.then(function(name) {
+						// check if the load corresponds to a babel plugin,
+						// otherwise a loading plugin could try to load itself
+						if (load.name !== name) {
+							return self.import(name);
```
matthewp: I could be wrong but I feel like this needs to be `self["import"]` for IE9 maybe? At least that was the case with IE8, maybe should check if it's ok to use `.import` in IE9.

m-mujica:
- [ ] loading plugins from node_modules 

* the name in the plugins array does not match the name in the node_modules folder, so steal fails to load it. Babel plugins have a convention for naming though, so I think we should be able to detect if the plugin name is a relative path, in which case we assume it's a custom plugin path relative to the `main` or take the plugin name and prefix it with `babel-plugin-`.

m-mujica:
@matthewp finally got the saucelabs build passing

matthewp:
1q3gh9j

Issue #1104, opened 2017-03-17
Meta deps not being normalized by npm extension


m-mujica:
https://github.com/m-mujica/steal-meta-deps-npm

```json
 "meta": {
    "justifiedGallery/src/js/justifiedGallery": {
      "format": "global",
        "deps": [
          "jquery",
          "justifiedGallery/src/less/justifiedGallery.less"
       ]
    }
}
```

I'm getting `GET http://localhost:8080/jquery.js 404 (Not Found)`, Steal should be looking for jquery in the node_modules folder.

m-mujica:
`justifiedGallery` does not define jquery as a dependency in its package.json.

Steal is normalising the `load.metadata.deps` using `justifiedGallery` as its parent module and since its package.json is missing the jquery dependency entry, the npm extension "thinks" it's a regular module. 

wondering if there is anything we can do in Steal to handle this, or if we should care about this at all. (cc: @matthewp)

We could modify the progressive loading guide to use the full path to node_modules folder instead @justinbmeyer 

```json
 "meta": {
    "justifiedGallery/src/js/justifiedGallery": {
      "format": "global",
        "deps": [
          "node_modules/jquery/dist/jquery.js",
          "justifiedGallery/src/less/justifiedGallery.less"
       ]
    }
}
```

thoughts?

matthewp:
I think the bug here is that it is using `justifiedGallery` as the parent. The parent should be the package where this configuration is defined.

m-mujica:
@matthewp gotcha, I'll look into it.

m-mujica:
@matthewp so the change should be around here:

https://github.com/stealjs/steal/blob/8c0c2239c54aba8c98677e126164d4d95f21e452/steal.js#L4002

I could pass `loader.configMain` instead of `load.name` as the parent. Is it safe to assume all module names in `load.metadata.deps` need to be normalized against the `configMain`?

m-mujica:
Also I wrote this test https://github.com/stealjs/steal/pull/1110

matthewp:
> @matthewp so the change should be around here:

We talked about this offline but for record, this is not the solution. The solution is to update npm-convert to handle meta.deps.

matthewp:
This is a dupe of https://github.com/stealjs/steal/issues/971

PR #1106, opened 2017-03-20
Progressive loading: use jquery's full path to node_modules folder

m-mujica:
Ref: steal/issues/1104

Based on #1105, actual change https://github.com/stealjs/steal/pull/1106/commits/406e34ca9edb937172249ba47509333eebd33c0a


pYr0x:
so we dont want do normalize the deps through npm ? like `jquery` https://github.com/stealjs/steal/issues/1104

m-mujica:
@pYr0x we do normalize the deps through npm, the problem with #1104 is that the "jquery's parent module, `justifiedGallery` doesn't list `jquery` as a dependency and the npm extension doesn't normalize the name correctly.

Not sure if this is a bug, will take a look at it in a little bit.

PR #1110, opened 2017-03-21
WIP: test for #1104

m-mujica:


matthewp:
https://github.com/stealjs/steal/blob/8c0c2239c54aba8c98677e126164d4d95f21e452/ext/npm-convert.js#L46

I think the solution is to, right around there, do conversions on meta.deps as well.

PR #1117, opened 2017-03-22
Allow loading packages that are not direct dependencies

### LINKED ISSUE #971: Unlisted sub-dependencies don't resolve with parent config
Here's a real world case I ran into:

jQuery UI doesn't have jQuery listed as a dependency, yet requires jQuery. The dependencies are then resolved incorrectly, failing to resolve using parent configuration. 

System._traceData.loads["jquery-ui@1.12.1#ui/widgets/autocomplete"].metadata.dependencies is resolving to "jquery" instead of "jquery@3.1.1/dist/jquery"

This problem can't be fixed with meta or map, as those aren't aware that jQuery is a npm dependency when tracing from within a package missing dependencies.

matthewp:
This change makes it possible so that you can load dependencies from
which you do not actual depend on.

Meaning if you have:

```js
require("jquery");
```

In your code, but your package does not directly depend on `jquery`,
   then it is possible you will load jquery from npm anyways; if jquery
   happens to exist within any parent package.

Closes #971

### DIFF
--- a/ext/npm-crawl.js
+++ b/ext/npm-crawl.js
@@ -429,7 +429,7 @@ var crawl = {
 	 * Find a package and its parents.
 	 * [package:{}, parent1, parent2, ...]
 	 * @param {Object} context
-	 * @parent {String} name the package name
+	 * @param {String} name the package name
 	 */
 	findPackageAndParents: function(context, name) {
 		return context.packageParents[name];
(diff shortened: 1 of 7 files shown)


ext/npm-crawl.js
```diff
@@ -408,6 +408,32 @@ var crawl = {
...
+			}
+			parents.push(pkg);
+		});
+	},
+	/**
+	 * @function findPackageAndParents
+	 * Find a package and its parents.
+	 * [package:{}, parent1, parent2, ...]
+	 * @param {Object} context
+	 * @parent {String} name the package name
```
m-mujica: `@param`

m-mujica:
just a minor thing, LGTM though.

PR #1122, opened 2017-03-23
Support environment dependant babel plugins

### LINKED ISSUE #1102: Support environment dependant babel plugins
https://github.com/stealjs/steal/pull/1100 adds the ability to load babel plugins not included in `babel-standalone`. 

Some babel plugins are meant to be run in certain environments, e.g https://github.com/istanbuljs/babel-plugin-istanbul uses `fs` and should not be loaded in the browser.

Babel supports an `env` option https://babeljs.io/docs/usage/babelrc/#env-option

- [x] Collect plugins set in the `env` object
- ~~[ ] Make steal set the BABEL_ENV somehow ?~~
- [x] Only import the plugins if `loder.isEnv(pluginEnv)` is true 

cc: @matthewp

m-mujica:
Closes #1102


matthewp:
1q3gh9j

PR #1123, opened 2017-03-24
Prevent System.ext from having a 'constructor'

### LINKED ISSUE #938: plugins() appends '!function Object() { [native code] }' to loads that end with '.constructor'
To fix this, change https://github.com/stealjs/steal/blob/master/src/system-extension-ext.js#L8 to:

```javascript
		loader.ext = Object.create(null);
```

and https://github.com/stealjs/steal/blob/master/src/config.js#L38 to:

```javascript
	System.ext = Object.create(null);
```

[`Object.create(null)` creates objects with no constructor or proto -- better for dictionaries than `{}`]

matthewp:
The System.ext configuration is used to support Steal's ability to load
plugins without the bang. Since it was previously a plain object, it had
a constructor property. This meant if you tried to do:
`require("./foo.constructor");` where this is actually
`foo.constructor.js`, it would think that it needed to use `ext`
configuration.

The fix is to use `Object.create(null)` which creates an object with no
properties. Closes #938


m-mujica:
@matthewp 👍 or 😐

matthewp:
Ah, you figured out what the bug was, thanks!

PR #1131, opened 2017-03-28
Add __esModule flag to modules transpiled on the browser

m-mujica:
With dev bundles is possible to load a part of the app already built while other modules are being transpiled on the fly, with this flag, transpiled amd modules will be able to load the modules transpiled on the browser.


matthewp:
LGTM 👍

matthewp:
Which bug does this fix?

m-mujica:
Closes https://github.com/stealjs/steal/issues/1136

PR #1137, opened 2017-03-29
Update babelOptions documentation

### LINKED ISSUE #1112: Update babelOptions documentation
The documentation is currently a little dated: http://stealjs.com/docs/config.babelOptions.html

I think we should

* [ ] - Remove stuff about Babel 5. Steal uses Babel 6 now.
* [ ] - Documentation special behavior on using custom plugins (like the rules around `./` and `babel-plugin-` and such).
* [ ] - The page shows "object" with no text inside. It should contain `@option`s for preset and plugin.

screen shot 2017-03-22 at 11 27 09 am

m-mujica:
Closes #1112

babel-options

babel-options-bottom

### DIFF
--- a/docs/config-babelOptions.md
+++ b/docs/config-babelOptions.md
@@ -1,61 +1,77 @@
-@property {Object} config.babelOptions babelOptions
+@typedef {Object} config.babelOptions babelOptions
 @parent StealJS.config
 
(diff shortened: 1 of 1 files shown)


m-mujica:
The example app code mentioned in the Generators section https://github.com/stealjs/steal-generators-example

PR #1141, opened 2017-04-02
Allow @/ to work the same as ~/

marshallswain:
This allows modules beginning with `@/` to resolve the same as `~/`.

Webpack projects use `@/` the same as our `~/` operator.   Webpack doesn't allow working in modlets, so I saw a good opportunity to promote Steal as a tool to use together with webpack.  I'm going to be posting an article in the Vue.js forums on how its preferable to use Steal during development because it allows working in modlets. (Vue.js's CLI builds webpack projects)  Making `@/` work is the last requirement to making the workflow perfect.

### DIFF
--- a/docs/scheme-tilde.md
+++ b/docs/scheme-tilde.md
@@ -1,4 +1,4 @@
-@property tilde ~
+@property tilde ~ or @
 @parent StealJS.schemes
 
 A lookup scheme that roots module lookup to your project's base folder, either your `steal.directories.lib` folder or the [config.baseURL].
@@ -7,9 +7,10 @@ This syntax is supported by all module formats.
 
 ## Use
 
-Prepend lookups with `~/` such as:
+Prepend lookups with `~/` or `@/` such as:
 
     var tabs = require("~/components/tabs/tabs");
+    var tabs = require("@/components/tabs/tabs");
 
 This will load the module from `BASE/components/tabs/tabs.js`. If your package.json has:
 
@@ -26,7 +27,7 @@ Then it will be loaded from `BASE/src/components/tabs/tabs.js`.
 
 ## Alternatives
 
-The **~** scheme is an alternative to using the package name for look up, such as:
+The **~** or **@** scheme is an alternative to using the package name for look up, such as:
 
      {
 	   "name": "app"
--- a/ext/npm-utils.js
+++ b/ext/npm-utils.js
@@ -554,7 +554,7 @@ var utils = {
 			return  path.substr(0,1) === ".";
 		},
 		startsWithTildeSlash: function( path ) {
-			return path.substr(0,2) === "~/";
+			return path.substr(0,2) === "~/" || path.substr(0,2) === "@/";
 		},
 		joinURIs: function(base, href) {
 			function removeDotSegments(input) {
--- a/test/npm/test.js
+++ b/test/npm/test.js
@@ -86,6 +86,19 @@ QUnit.test("meta", function(assert) {
 		});
 });
 
+QUnit.test("@/ works the same as ~/", function(assert) {
+	var done = assert.async();
+
+	GlobalSystem["import"]("@/meta")
+		.then(function(meta) {
+			assert.equal(meta, "123", "got 123");
+		})
+		.then(done, function(err) {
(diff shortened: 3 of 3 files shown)


matthewp:
Cool, I wasn't aware of this convention. Do you have any example project or code I could look at to see this? Is this specifically a thing within the Vue community?

marshallswain:
I thought this was a webpack thing in general, but it looks like it's enabled in the config:

```js
resolve: {
    alias: {
      '@': resolve('src')
    }
  },
```

I suppose that could easily be changed to a tilde, but I'd like to push Steal as a tool they could add to improve their existing apps & development workflow.  Would it be possible to use `map` in the Steal config to do the same thing?

matthewp:
Map would work if not for directories.lib 🙁

What if we made the tilde behavior configurable?  Conceptually this feature is trying to mirror the `~` alias in Unix operating systems. `~` is an alias for `$HOME`.  What if we called it `homeAlias`?

```json
"steal": {
  "homeAlias": "@"
}
```

What do you think?  cc @m-mujica

leoj3n:
I think `rootAlias` might be better to differentiate from, like you said, the home dir of the OS.

marshallswain:
Configurable is even better! Either recommendation works for me.

matthewp:
I don't mind rootAlias except that it sort of conflicts with `root` (which is the same as `baseUrl`).

m-mujica:
@matthewp the webpack version allows you to alias any piece of a module name that's constantly used. 

https://webpack.js.org/configuration/resolve/#resolve-alias

That sounds like a nicer feature to me, then the "root/home" alias is just a special use case of the feature.

marshallswain:
Maybe we can make `homeAlias` work before we look at a bigger scope.  I'd like to get writing about this stuff while I'm still motivated.  ;)

leoj3n:
Seems like you could make the API future proof while still only supporting
home/root at first?

matthewp:
@m-mujica `alias` looks like its the same as our map config, right? If so, we can't use it in this case (unfortunately) because when using the npm plugin every module name gets turned into an npm-style name and directory map config doesn't work with the npm plugin 😢.

That's why packageName / tilde is sort of a special-case and we don't use map config. If we [deprecate directories.lib](https://github.com/stealjs/rfcs/issues/13) then we can use map config instead of something like this.

But since we can't do that until 2.0, I think it's better to extend the existing option just a bit.

If we all agree @marshallswain @m-mujica @leoj3n then I think we should go with `homeAlias`.

m-mujica:
70a8541370bfc969ddd3058d705fc64d

m-mujica:
j/k, sounds good to me.

marshallswain:
pipe down, grumpy cat!

matthewp:
Cool, so since we have consensus I think we should go ahead with `homeAlias`.

@marshallswain I would add a function in [here](https://github.com/stealjs/steal-npm/blob/216463c552cabe7d9dad86d754da784b5dea3d78/npm-utils.js#L336) called `homeAlias` that you can use like `utils.pkg.homeAlias(context)` that should return `context.loader.homeAlias || '~'`. Something like that.  Then use this function in the startsWithTildeSlash function (I would rename this function to `isInHomeDir` maybe).

docs/scheme-tilde.md
```diff
@@ -1,4 +1,4 @@
-@property tilde ~
+@property tilde ~ or @
```
matthewp: Let's call this "Home Alias" as the title. Leave the name as `tilde` for existing SEO.

ext/npm-utils.js
```diff
@@ -554,7 +554,7 @@ var utils = {
 			return  path.substr(0,1) === ".";
 		},
 		startsWithTildeSlash: function( path ) {
-			return path.substr(0,2) === "~/";
+			return path.substr(0,2) === "~/" || path.substr(0,2) === "@/";
```
matthewp: cache the value of `path.substr(0, 2)` in a variable rather than performing it twice.

marshallswain:
Closing in favor of https://github.com/stealjs/steal/pull/1178

PR #1144, opened 2017-04-04
JSON extension should not parse css files

### LINKED ISSUE #1092: CSS is parsed as JSON if file starts with an attribute selector
@phillipskevin commented on [Thu Mar 09 2017](https://github.com/stealjs/steal-css/issues/32)

Reported [on gitter](https://gitter.im/stealjs/steal?at=58c01584e961e53c7fb7f44e). If a CSS file starts with an attribute selector (like `[name="test"]`), steal-css will try to parse it as JSON:

image

---

@matthewp commented on [Thu Mar 09 2017](https://github.com/stealjs/steal-css/issues/32#issuecomment-285380881)

Most likely a bug in Steal itself, the json extension would be my guess.

m-mujica:
Closes #1092

@matthewp what do you think about this one? I thought about checking if `load.metada.buildType` wasn't `css` but that seemed brittle to me.

The thing about using this helper is that before, json files with comments on it (which technically are not json files) didn't fail to load, but now they will fail. 

I had to tweak one of the tests to have a "malformed" json instead of an invalid one.


matthewp:
json files can't have comments, i think its ok to break already-broken json :)

main.js
```diff
@@ -796,10 +796,20 @@ addStealExtension(function applyTraceExtension(loader) {
...
-  var jsonTest = /^[\s\n\r]*[\{\[]/;
   var jsonExt = /\.json$/i;
   var jsExt = /\.js$/i;
 
+  // taken from prototypejs
+  // https://github.com/sstephenson/prototype/blob/master/src/prototype/lang/string.js#L682-L706
+  function isJSON(str) {
+    if (!str) return false;
+
+    str = str.replace(/\\(?:["\\\/bfnrt]|u[0-9a-fA-F]{4})/g, '@');
```
matthewp: this might be dated information... but it used to be true that defining a regex inside a function caused it to be created each time. I'm not sure if they are treated as constants by modern js engines or not (someone to look into perhaps).  I usually write in this form for this kind of thing:

```js
var isJSON = (function(){
  var firstRegExp = /.../;

  return function(str){
  };
})();
```

m-mujica: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp)

```
The literal notation provides compilation of the regular expression when the expression is evaluated. Use literal notation when the regular expression will remain constant. For example, if you use literal notation to construct a regular expression used in a loop, the regular expression won't be recompiled on each iteration.
```

also http://stackoverflow.com/questions/8814009/how-often-does-javascript-recompile-regex-literals-in-functions

I think we are good there, want me to move them outside anyway?

matthewp: Nah, sounds good to me.

PR #1161, opened 2017-04-10
Improve CJS module format detection

### LINKED ISSUE #880: CommonJS format not detect when using Object.defineProperty on module
For example:

```js
Object.defineProperty(module, 'exports', {

});
```

Related to https://github.com/stealjs/steal/issues/375

m-mujica:
Enhances RegEx to detect modules using Object.defineProperty to set `module.exports`

Closes #880


matthewp:
:+1:

PR #1164, opened 2017-04-12
Log an error when unable to connect to ws server

### LINKED ISSUE #1163: Let the user know when live-reload is unable to connect
If there is a connection issue in live-reload it should throw an "error" event. We won't know the reason for the error but we should at least tell the user that there is a problem.

matthewp:
Closes #1163

Also enables the live-reload unit tests in CI.

### DIFF
--- a/ext/live-reload.js
+++ b/ext/live-reload.js
@@ -282,6 +282,11 @@ function setup(){
 		reloadAll(moduleName);
 	};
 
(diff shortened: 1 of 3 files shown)


m-mujica:
🚢 🇮🇹

m-mujica:
f0ctsxp

PR #1168, opened 2017-04-13
Update readme

leoj3n:
I don't remember when/where we talked about this, but I just wanted to open a PR so it doesn't get lost/forgotten. I think @matthewp had some reservations about updating the README, and mentioned something about list of plugins that the org maintains.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,14 +7,105 @@
 
 [![Sauce Test Status](https://saucelabs.com/browser-matrix/matthewphillips.svg)](https://saucelabs.com/u/matthewphillips)
 
-`steal` is a JavaScript module loader. `steal` supports ES6, AMD, CommonJS, LESS, and 
-CSS modules. It makes development workflows simple.
+In addition to a collection of plugins, __StealJS__ is comprised of two main components:
 
-It is one half of the [StealJS](http://stealjs.com) project.  It's other half,
-[steal-tools](https://github.com/stealjs/steal-tools) provides high performance builds.
+  - __`steal`__: an extensible, universal module loader.
+  - __`steal-tools`__: utilities for building, transforming, and exporting module formats.
 
-For more information on `steal`, visit [stealjs.com](http://stealjs.com).
+This is the `steal` repository. For the `tools`, see <https://github.com/stealjs/steal-tools>.
 
-For information on contributing and developing `steal`, go [here](http://stealjs.com/docs/guides.Contributing.html).
+## Module Loading with `steal`
 
+`steal` is unique because it can load JavaScript modules defined in ES6, AMD, and CommonJS formats (unlike most other module loaders, which only support one of these formats at a time).
+
+In JavaScript, the word "modules" refers to small units of independent, reusable code. They are the foundation of many JavaScript design patterns, and can look like this in ES6:
+
+```js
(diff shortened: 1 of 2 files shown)


matthewp:
I'm ok with updating the readme. I'll review this.

README.md
```diff
@@ -7,14 +7,105 @@
 
 [![Sauce Test Status](https://saucelabs.com/browser-matrix/matthewphillips.svg)](https://saucelabs.com/u/matthewphillips)
 
-`steal` is a JavaScript module loader. `steal` supports ES6, AMD, CommonJS, LESS, and 
-CSS modules. It makes development workflows simple.
+In addition to a collection of plugins, __StealJS__ is comprised of two main components:
 
-It is one half of the [StealJS](http://stealjs.com) project.  It's other half,
-[steal-tools](https://github.com/stealjs/steal-tools) provides high performance builds.
+  - **Steal**: an extensible, universal module loader.
```
matthewp: we use lower-case `steal`

README.md
```diff
@@ -7,14 +7,105 @@
...
+  - **Steal Tools**: utilities for building, transforming, and exporting module formats.
```
matthewp: Lower case `steal-tools`

README.md
```diff
@@ -7,14 +7,105 @@
...
+  - [Less](https://github.com/stealjs/steal-less)
+  - [Conditional](https://github.com/stealjs/steal-conditional)
+  - [Mocha](https://github.com/stealjs/steal-mocha)
+  - [QUnit](https://github.com/stealjs/steal-qunit)
+
+For example, the Less plugin allows Less files to be loaded similarly to JavaScript modules:
+
+```js
+// ES6
+import "style.less!";
```
matthewp: The bang is not necessary so we omit it now.

matthewp:
Just a couple of things, thanks!

leoj3n:
@matthewp Did you also comb through the code examples? I took my best stab at it, but am not entirely sure I hit the mark, as I've not used most of these module formats outside of StealJS.

matthewp:
Yep, looks good, thanks!

m-mujica:
I think this PR would be way more awesome if it were a single atomic change (single commit) and speaking about commits, using a guideline like https://github.com/erlang/otp/wiki/Writing-good-commit-messages would make it even more amazing.

We don't have an official commit message guideline, but most of them look as nice as this one 

https://github.com/stealjs/steal/commit/f27647691fd862c312847ce069db5ec3e7604dad

just sayin....

m-mujica:
Also, I think you need to rebase against master @leoj3n, not quite sure what's going on with Saucelabs. I've tried rebuilding your PR a couple of times and got the same errors.

leoj3n:
I just assumed you would squash on merge. You can do it through the web UI
even. I'll rebase on my end though.

leoj3n:
Normally I force push to my PRs, but I got yelled at for that recently because there were unsresolved Q/As (although I'm still not 100% sure if that actually bumps out the Q/As or not).

matthewp:
Moved this PR over to https://github.com/stealjs/steal/pull/1260 which is merged with master so tests should pass. Thanks for the contribution!

PR #1204, opened 2017-05-18
Use double quotes in html attributes

m-mujica:
Part of stealjs/issues/46


m-mujica:
hmm I got the same error about the environment being unavailable.

screen shot 2017-05-18 at 09 50 20

I don't think we are supposed to update the appium version every time they change the default 😓  

screen shot 2017-05-18 at 09 50 07

I'll try just with a rebuild. We'll probably need to look into it. @matthewp

PR #1214, opened 2017-06-09
Make the live-reload extension cloneable

### LINKED ISSUE #1213: live-reload extension is not cloneable
The live-reload extension doesn't get carried over to cloned loader instances. This causes problems when modules import live-reload inside of a cloned loader, the extension's normalization isn't used so the wrong value of `live-reload` in returned.

matthewp:
This makes it so that the live-reload extension is cloneable, and
therefore doesn't cause errors when used with steal-clone in tests.

Closes #1213

### DIFF
--- a/ext/live-reload.js
+++ b/ext/live-reload.js
@@ -1,10 +1,6 @@
 var loader = require("@loader");
 var steal = require("@steal");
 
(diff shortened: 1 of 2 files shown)


matthewp:
Going to wait to merge/release this until monday to avoid weekend disasters.

Issue #1221, opened 2017-07-02
steal-clone path errors


justinbmeyer:
> Originally, I thought this had to do with clone happening within dynamic imports.  As demonstrated by #1222, I think there's just an error in `steal-clone` and how it determines paths.

Demonstrated by #1222.

This problem can be seen by cloning the `updateDeep` branch of canjs and changing `test/test.js` page to:

```js
var order = Promise.resolve();

if (!System.isEnv('production')) {
	order = order.then(function(){
		return System.import('can-simple-map/can-simple-map_test');
	});
}
if (!System.isEnv('production')) {
	order = order.then(function(){
		return System.import('can-cid/test/test');
	});
}
```

Both of these tests use `steal-clone`.  If only one `System.import` is present, everything works. 
 The error is that the `can-cid` test will not load:

![_canjs_tests_and_slack_-_bitovi](https://user-images.githubusercontent.com/78602/27766590-d92ff4e8-5e9a-11e7-9122-e4f1a48432a8.png)

I'm working on creating a smaller test right now that demonstrates this problem.

justinbmeyer:
This also fails if `@loader` is used everywhere:

```js
var loader = require("@loader");

var order = Promise.resolve();

if (!System.isEnv('production')) {
	order = order.then(function(){
		return loader.import('can-simple-map/can-simple-map_test');
	});
}
if (!System.isEnv('production')) {
	order = order.then(function(){
		return loader.import('can-cid/test/test');
	});
}
```

justinbmeyer:
I wasn't easily able to create this problem in `steal`.   But I've reduced the `canjs` test to:

```js
// test/test.js
var loader = require("@loader");
loader.import('can-simple-map/can-simple-map_test');
loader.import('can-cid/test/test');
```

```js
// node_modules/can-simple-map/can-simple-map_test
var clone = require('steal-clone');
```

```js
// node_modules/can-cid/test/test
require('can-cid/can-cid_test'); // relative here breaks too
```

```js
// node_modules/can-cid/can-cid_test.js
var clone = require('steal-clone');
clone().import('can-cid')
```

PR #1222, opened 2017-07-02
Fix bug with supporting multiple contextual modules of one type

justinbmeyer:
This fixes and adds tests for a problem with multiple clones (#1221).  The fix is prevent only a single contextual loaded module from being created.  Previously, a new contextual module would be defined only if the following check passed:

```js
if(!loader.has(name)) {
```

As name was always `steal-clone`, this would only allow one contextual module.

I changed to to:

```
if(!loader.has(localName)) {
```

This will only allow one contextual module of that `name+parentName` combination to be loaded.

~~Test are not passing for me locally ... I assume this is because I'm on node 5.~~ Because I didn't have bower installed.


justinbmeyer:
I just pushed another update showing that this error doesn't have to happen within dynamic imports.  I changed `test.js` to:

```
require("steal/test/clone-within-dynamic-imports/first-test");
require("steal/test/clone-within-dynamic-imports/second-test");
```

justinbmeyer:
Odd part is that normalize of `steal-clone` within`steal@1.5.4#test/clone-within-dynamic-imports/first-test` produces:

`steal-clone/steal@1.5.4#test/clone-within-dynamic-imports/first-test`

justinbmeyer:
Some more info ...  the contextual modules hook is creating the name `steal-clone/steal@1.5.4#test/clone-within-dynamic-imports/first-test` [here](https://github.com/stealjs/steal/blob/71ef85c8b2c9dd9d3683766f57710c75701b1d18/src/system-extension-contextual.js#L18).  Is this expected?  

It seems its trying to automatically create this module, but it's not being found later.  Still investigating why ...

justinbmeyer:
Ah, I think the problem is that only one `steal-clone` contextual module can be loaded:

```js
if(!loader.has(name)) {
```

needs to be changed to:

```js
if(!loader.has(localName)) {
```

justinbmeyer:
Can confirm, this fixes it.  Going to integrate the test and hopefully make a `.pre` patch release I can point CanJS to.  Once @matthewp or @m-mujica give this a thumbs up, I'll make a real release.

matthewp:
Great! Thanks.

matthewp:
Released as https://github.com/stealjs/steal/releases/tag/v1.5.5

PR #1225, opened 2017-07-19
Continue crawling when a incompatible parent is found

### LINKED ISSUE #1224: Nested dependency "ms" not loading correctly

Sometimes when served from localhost, always served from Testee on Firefox.

__Description:__

*Note: this is in the context of my attempt to get `can-fixture-socket`'s build passing again.*

I am loading the library `ms` which is a dependency of `debug` which is a dependency of a lot of other projects. Two versions of `ms` are needed looking at `npm ls ms`, versions ^2.0.0 and ^0.7.1. When I load a few dependencies in Steal I get the following errors:

In Firefox (which was run with Testee):
screen shot 2017-07-12 at 3 57 39 pm

In Chrome (serving from http-server):
screen shot 2017-07-13 at 9 42 30 am

If not run in Testee, this issue is intermittent even if the code and `node_modules` have not changed. Somehow something is not deterministic.

__Steps to reproduce:__

I created a [slimmed down version of `can-fixture-socket`](https://ufile.io/pv22x) however I was unable to reduce the number of dependencies as this seems, to me, to be a perfect storm issue.

`npm install && npm test` in the project folder to see the issue. If you don't re-try a few times, but this fails on my local and on CI with `can-fixture-socket`.

__Expected results:__

I expect the `ms` module to be found and Steal allow the `can-fixture-socket` test suite to run.

__Actual results:__

The `ms` module is not found and Steal prevents the `can-fixture-socket` test suite from running.

__Environment:__

| Steal version              | ^1.2.8
| Steal-tools version    | ^1.1.2
| node -v                       | 8.1.2
| npm -v                        | 4.6.1
| Browser                      | Firefox/Chrome
| Operating system      | Mac OS / Travis's

m-mujica:
The logic to get the `package.json` was stopping when a parent with an incompatible version was found, this causes some issues where steal would fetch the file in an incorrect location. Instead, the crawling should continue looking for parents where the file might be located.

Closes #1224

### DIFF
--- a/ext/npm-crawl.js
+++ b/ext/npm-crawl.js
@@ -347,8 +347,6 @@ var crawl = {
 			if(parentPkg && SemVer.valid(parentPkg.version)) {
 				if(SemVer.satisfies(parentPkg.version, childPkg.version)) {
 					return parentPkg.fileUrl;
-				} else {
-					return curAddress;
 				}
 			}
 			parentAddress = utils.path.parentNodeModuleAddress(packageAddress);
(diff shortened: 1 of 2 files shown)


matthewp:
lgtm

PR #1229, opened 2017-07-25
Better error message for possible unsaved packages

### LINKED ISSUE #1205: Warning for unresolved packages not installed as dep or devDep
If someone installs an NPM package without saving it to the package.json, as a fall-through warning message we can say "It looks like "{packageName}" exists in node_modules but is not in your package.json's dependencies or devDependencies, please add it if this is your intention." Something to that effect.

m-mujica:
Closes #1205

screen shot 2017-07-25 at 16 10 17

is rather verbose but better than the alternative I suppose.


matthewp:
Do you have any idea why it prints twice? Is that just how console.error works?

m-mujica:
@matthewp it turned out to be that we were passing `error.stack` as a second argument.

Safari's console output is even worse (WAT)

screen shot 2017-07-25 at 17 35 18

I changed it so only error is passed and tried in Chrome, Safari and Firefox, looks way better and the stack is still there.

screen shot 2017-07-25 at 17 38 03

screen shot 2017-07-25 at 17 22 54

screen shot 2017-07-25 at 17 40 46

What do you think?

matthewp:
Awesome!

PR #1236, opened 2017-08-01
Fixes bug loading peer deps that are dev deps too

### LINKED ISSUE #1228: can not load JSS with steal

- [x] Always

__Description:__
tried to load JSS https://github.com/cssinjs/jss
```
import jss from 'jss';
import preset from 'jss-preset-default';
```
in my index.js file.

__Steps to reproduce:__

1. clone repo https://github.com/pYr0x/jss-steal-app/tree/jss-bug
2. npm install
3. open dev.html

__Expected results:__
all modules are loaded.

__Actual results:__
there is a problem with the module `jss-global`. the package.json is loaded correctly. this module tries to load `jss` in `node_modules/jss-global/lib/index.js` line 13. 
`jss-global` defines `jss` as a peerDependency (see the package.json of jss-global). but i get an error that steal can not load the package.json of `jss`
`http://localhost:63342/jss-steal-app/node_modules/jss-global/node_modules/jss/package.json`
it seems that steal tries to load `jss` within `jss-global` folder. but thats not correct.

does steal not correctly handle peerDependencies?

__Environment:__

| Steal version              | 1.5.5
| Steal-tools version    | 1.7.0
| node -v                       | 6
| npm -v                        | 3.10
| Browser                      | chrome
| Operating system      | mac

m-mujica:
Closes #1228 

So, we already faced this problem before... I remember skipping this test while I was migrating steal-npm but forgot to file an issue to check why it was failing.

The issue is caused by the following scenario (in the test):

// package.json
```json
dependencies: {
   "@cycle/core": "^6.0.0",
  "rx": "^4.0.7"
}
```

// cycle/core's package.json
```json
devDependencies: {
  "rx": "4.0.6"
},
peerDependencies: {
  "rx": "*"
}
```

When the npm extension is normalizing `@cycle/core ->rx`, it grabs the version inside `devDependencies` which is not compatible with the one in the top package.json.

https://github.com/stealjs/steal/blob/d0e96bae5c672bfd859e9e13eef9e2e53f52163b/ext/npm-extension.js#L123

That line is the issue since it uses `getFullDependencyMap` which uses the packages inside `devDependencies`; I'm not sure what should be the condition to ignore dev dependencies though. cc: @matthewp

### DIFF
--- a/ext/npm-crawl.js
+++ b/ext/npm-crawl.js
@@ -267,9 +267,13 @@ var crawl = {
 	 */
 	getFullDependencyMap: function(loader, packageJSON, isRoot){
 		var deps = crawl.getDependencyMap(loader, packageJSON, isRoot);
-		addDeps(packageJSON, packageJSON.devDependencies || {}, deps,
-			"devDependencies");
-		return deps;
+
+		return addMissingDeps(
+			packageJSON,
+			packageJSON.devDependencies || {},
+			deps,
+			"devDependencies"
+		);
 	},
 	getPlugins: function(packageJSON, deps) {
 		var config = utils.pkg.config(packageJSON) || {};
@@ -488,6 +492,22 @@ function addDeps(packageJSON, dependencies, deps, type, defProps){
 			deps[name] = val;
 		}
 	}
+}
+
+/**
+ * Same as `addDeps` but it does not override dependencies already set
+ */
+function addMissingDeps(packageJson, dependencies, deps, type, defProps) {
+	var without = {};
+
+	for (var name in dependencies) {
+		if (!deps[name]) {
+			without[name] = dependencies[name];
+		}
+	}
+
+	addDeps(packageJson, without, deps, type, defProps);
+	return deps;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


m-mujica:
Now, #1228 is same same different but still same.... The interesting bit about #1228 is that both devDependency/peerDependency point to the same version 

https://github.com/cssinjs/jss-preset-default/blob/master/package.json#L43-L48

In this case, steal loads the top level version which is 8.1.0, then when it tries to load the package of 8.0.0 it hits this code 

https://github.com/stealjs/steal/blob/d0e96bae5c672bfd859e9e13eef9e2e53f52163b/ext/npm-crawl.js#L568

and 

```javascript
// pkg.version           -> 8.1.0
// requestedVersion -> 8.0.0
SemVer.satisfies(pkg.version, requestedVersion) -> returns false
```

When this function is called we only have the package version values with no carets at all, so unless those version are exactly the same it will always detect them as incompatible I guess.

... thinking...

m-mujica:
@pYr0x I made an example app using this stealjs branch, jss is loading fine now.

jss

💥

m-mujica:
@matthewp care if I merge this one?

matthewp:
Nope, looks good to me.

PR #1239, opened 2017-08-09
Fix steal-module script extension

### LINKED ISSUE #1185: script type="steal-module" doesn't work
It looks like it has to be text/steal-module

m-mujica:
- Fix indentation
- Get tests running
- Make it so type="steal-module" works

Closes #1185


matthewp:
👍

PR #1245, opened 2017-08-10
Register full npm name of named AMD modules

### LINKED ISSUE #1237: Unable to load named AMD module with dependencies
@imaustink found this one, the problem is very strange. I was able to replicate with the following code.

```javascript
// index.js 
var foo = require("./foo");
console.log(foo);
```

```javascript
// foo.js
define("foo", ["./bar"], function(bar) {
  return "foo";
});
```

Basically if main requires an amd module with deps, boom, it throws

screen shot 2017-08-03 at 18 09 26

:/

m-mujica:
Closes #1237


test/npm/import_test.js
```diff
@@ -461,6 +461,79 @@ QUnit.test("importing a package with an unsaved dependency", function(assert) {
...
+			{
+				name: "foo",
+				version: "1.0.0",
+				main: "main.js"
+			}
+		])
+		.withModule("foo@1.0.0#bar", "module.exports = 'bar';")
+		.withModule(
+			"foo@1.0.0#main",
+			"def" + "ine('foo', ['./bar'], function(bar) { return bar; })"
```
m-mujica: @matthewp this is one question I won't ask you again, never.

m-mujica:
I put together a quick example loading sinon which triggered this issue initially

https://github.com/m-mujica/steal-sinon-example

screen shot 2017-08-10 at 16 25 44

and it works fine, a little dramatic but works :)

m-mujica:
@matthewp can I merge this one?

matthewp:
I made a comment but didn't submit the review so it didn't show it to you.

ext/npm-extension.js
```diff
@@ -258,6 +258,11 @@ exports.addExtension = function(System){
 	System.locate = function(load){
 		var parsedModuleName = utils.moduleName.parse(load.name),
 			loader = this;
+
+		if (load.metadata) {
+			load.metadata.parsedModuleName = parsedModuleName;
```
m-mujica: The test below fails without the conditional, `metadata` is not set for some reason. Want me to figure out why? Should `metadata` always be available at that point?

https://github.com/stealjs/steal/blob/915c636bef2018e7d3a30e2aaf8855b4c554c3aa/test/npm/test.js#L165-L185

matthewp: I'm pretty sure it should. search for `createLoad` I think. I think it creates the metadata object.  Could be the helpers stuff maybe.

m-mujica: Arggghh, I totally missed that the test is calling `locate` straight passing a partial load object 

```js
```

PR #1257, opened 2017-08-24
Fix typo "load import" in quick-start.md

leoj3n:
image


matthewp:
Thanks!  Looks like saucelabs were down, so I'm rerunning the tests.

PR #1271, opened 2017-09-18
Apply mappings when left-side is not a pkg but right-side is.

### LINKED ISSUE #1208: npm config `map` is not being set properly

- [x] Always
- [ ] Sometimes
- [ ] Rarely
- [ ] Unable
- [ ] I didn’t try

__Description:__
Steal config `map` does not seem to be being set properly.

__Steps to reproduce:__

> 1. Include a link to your project (if possible)  

Here is a gist you can clone to show the issue.
https://gist.github.com/BigAB/ae311de12272af411969f477e6f8c1a5

`preact-compat` should be a drop in replacement for `react` and `react-dom` and so map should swap them out fine. But in the gist, if you serve it you can see steal is still trying to import `react`.

The relevant part of the gist is mostly this in `package.json:
```
  "steal": {
    "map": {
      "react": "preact-compat",
      "react-dom": "preact-compat"
    }
  }
```

> 2. Detail the exact steps taken to produce the problem  

Clone the gist. Run `npm install`. Serve the root. See in the console that react is trying to load and `System.map` only has `@loader/@laoder` and `@steal/@steal` keys.

screen shot 2017-06-05 at 9 45 50 am

> 3. Include a gif if applicable

steal-config-map-issue

__Expected results:__

I expect everything to work, for every call to `import React from 'react` would return the `preact-compat` module instead

__Actual results:__
Steal fails to load `react` because it isn't there, and the config in `npm` does not even appear to be merged into `System.map`

__Environment:__

| Steal version              | 1.5.2
| Steal-tools version    | 1.3.5
| node -v                       | v8.0.0
| npm -v                        | 5.0.2
| Browser                      | Chrome 58.0.3029.110
| Operating system      | macOS Sierra 10.12.5

m-mujica:
Closes #1208

The app in the original issue runs ok with this branch https://github.com/m-mujica/steal-preact-compat-map

google chrome

### DIFF
--- a/ext/npm-convert.js
+++ b/ext/npm-convert.js
@@ -122,6 +122,10 @@ function convertPropertyNamesAndValues (context, pkg, map, root, waiting) {
 		if(typeof name !== 'undefined' && typeof val !== 'undefined') {
 			clone[name] = val;
 		}
+		// keep map entry if the key isn't a package but value might
+		if (name && typeof val === "undefined") {
+			clone[name] = map[property];
+		}
 	}
 	return clone;
 }
--- a/ext/npm-extension.js
+++ b/ext/npm-extension.js
@@ -214,8 +214,24 @@ exports.addExtension = function(System){
 					return oldNormalize.call(this, refPkg.browser[name], parentName,
 											 parentAddress, pluginNormalize);
 				}
-				return oldNormalize.call(this, name, parentName, parentAddress,
-										 pluginNormalize);
+				// Apply mappings, if they exist in the refPkg
+				var steal = utils.pkg.config(refPkg);
+				if (steal && steal.map && typeof steal.map[name] === "string") {
+					return loader.normalize(
+						steal.map[name],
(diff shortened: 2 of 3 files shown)


ext/npm-convert.js
```diff
@@ -122,6 +122,10 @@ function convertPropertyNamesAndValues (context, pkg, map, root, waiting) {
 		if(typeof name !== 'undefined' && typeof val !== 'undefined') {
 			clone[name] = val;
 		}
+		// keep map entry if the key isn't a package but value might
+		if (name && typeof val === "undefined") {
```
m-mujica: @matthewp is this a naive "package" detection? I couldn't think of a different way of "knowing" the key/val is a package without a full npm name which we won't have in this case.

matthewp: Yeah, I think this is fine.

matthewp:
lgtm

PR #1273, opened 2017-09-21
Warn if a module is loaded at the same path twice

### LINKED ISSUE #1264: Warn if someone loaded a module at the same path twice
Today I had someone import `models/todo` and `./models/todo`. This caused a situation where CanJS's real-time didn't seem to work.  This was because one part of the page was using one `Todo` export and other was using a different `Todo` export.  It took me a while to figure this was because there were two `Todo` types.

I'd like to propose warning someone if a file at a path has been loaded twice.

This can be done by keeping a side table storing paths to module names.  If two modules are added to this table, we should warn.

This warning would have made the problem immediately obvious.  

I think there are very few reasons the same module might want to be executed twice.  I remember @matthewp talking about this.  I don't think these cases are common enough where a warning wouldn't be helpful for a majority of users.

m-mujica:
Closes #1264

screen shot 2017-09-21 at 16 33 43


m-mujica:
- The extension is pretty much a no-op is `console.warn` is not available 
- The extension is only added to the unminified steal versions.

Any feedback about the message?

I feel it could use some explanation on why multiple instantiation might be bad, but the error is already kinda long.

cc: @justinbmeyer @matthewp

src/system-extension-module-loaded-twice.js
```diff
@@ -0,0 +1,48 @@
...
+	loader._instantiatedModules = loader._instantiatedModules || {};
+
+	loader.instantiate = function(load) {
+		var instantiated = loader._instantiatedModules;
+
+		if (warn && instantiated[load.address]) {
+			var loads = (loader._traceData && loader._traceData.loads) || {};
+			var map = (loader._traceData && loader._traceData.parentMap) || {};
+
+			var parents = (map[load.name] ? Object.keys(map[load.name]) : [])
```
m-mujica: @matthewp is it safe to use Object.keys? I think it's in IE9+

src/system-extension-module-loaded-twice.js
```diff
@@ -0,0 +1,48 @@
...
+				})
+				.join("\n");
+
+			warn(
+				[
+					"The module with address " + load.address +
+						" is being instantiated twice",
+					"This happens when module identifiers normalize to different module names.\n",
+					"HINT: Import the module using the ~/[modulePath] identifier" +
+						(parents ? " in " : ""),
```
m-mujica: Not sure if we should be defensive about modules without registered parents, I was thinking about main maybe.... Not sure.

m-mujica:
@matthewp this is ready to review now. I was doing `console.warn.bind(console)` and `console` is weird in IE9. ಠ_ಠ

matthewp:
👍

PR #1283, opened 2017-10-03
Use Node.js native promises

### LINKED ISSUE #1282: Use Node.js native promises
Node.js has native support since ~0.12 http://node.green/#ES2015-built-ins-Promise

m-mujica:
Closes #1282


m-mujica:
just a random thing I noticed debugging a build, also that `loader-no-promise` seems to be an old artifact from the original PR where the `sans-promises` files were introduced.

Using native promises helps with stack traces a little bit:

a) using polyfill
screen shot 2017-10-03 at 17 07 46

b) with native promises
screen shot 2017-10-03 at 17 08 43

matthewp:
Does this run only in Node? Browser code will still get the polyfill, right?

m-mujica:
> Does this run only in Node? Browser code will still get the polyfill, right?

Yup, that's the only place where that require happens and it only happens in Node.

https://github.com/stealjs/steal/blob/42e8ba8037a97198ea871eda796f963f59636640/main.js#L3888-L3889

PR #1296, opened 2017-11-08
Fix issue breaking live-reload when used with dynamic-import

imaustink:
Closes canjs/can-view-parser#75


matthewp:
This is the cause of the closing tag warnings?

imaustink:
@matthewp nope. I Referenced the wrong issue by accident. I updated the reference and reopened that issue.

PR #1299, opened 2017-11-09
Prevent false positive 'loaded twice' warnings

### LINKED ISSUE #1285: Module loaded twice warning is conducive to false positives
Unfortunately it appears that this warning message can often cause false positives. Here's one example:

```html
<script src="./node_modules/steal/steal.js"></script>
```

Will load the `steal.loader.configMain` as `package.json!npm`. User code might also have:

```js
var pkg = require("./package.json");
```

Thus, the package.json is loaded twice, but for two different reasons. This is valid to do.

This made me realize that loading files multiple times is a feature of steal, not a bug!

I'm not sure of a good solution here that preserves the warning without flooding users consoles with false positive messages.

The best solution I can come up with at the moment is to use `steal.dev.warn` and set the default logLevel to 2, so this warning would only be shown if users changed the log level to 0 or 1. cc-ing @justinbmeyer @m-mujica for ideas.

matthewp:
__Do not merge yet, discussing__

This changes how the false positive warnings work, requiring you to
*explicitly* set the log level to `0`. This means you now need to do:

```html
<script src="./node_modules/steal/steal.js" log-level="0"></script>
```

To see these warnings. This was the only option I could come up with,
   given that this type of message is prone to false-positives (because
 steal's registry is keyed by moduleName, not address).

For 2.0 I want to fix this log level stuff. Currently it defaults to 0.
It should default to 'warn' level or maybe even error.

Fixes #1285


justinbmeyer:
I'd like to discuss this at the contributor meeting.  Having experienced the problem that these warnings fix, I'm not sure hiding them is a better solution.

matthewp:
Ok, we can discuss then. I've had at least a half dozen people ask about this particular problem. I think you're more tolerable of warning messages than most, many people see them as errors. Too many false positive warnings and you start to ignore the *real* warnings. Some have reverted back to older versions of steal to prevent these warnings.

The only other solution that I can think of is to not use moduleNames as the registry keys any more.

matthewp:
Closing in favor of https://github.com/stealjs/steal/pull/1304 . Will reopen if we find more scenarios of false positives that cannot be fixed.

PR #1302, opened 2017-11-10
Add ability to define implicit dependencies using meta deps

### LINKED ISSUE #1297: Feature - ability to define implicit dependencies using meta deps
We are using `semantic-ui-react` and its sister `semantic-ui-css`. Each component depends on one or more CSS files, and we are wanting to do something like this so that developers can easily use components without thinking about the other dependencies:

```js
{
  map: {
    'semantic-ui-react/dropdown': 'semantic-ui-react/dist/es/modules/Dropdown/DropDown'
  },
  meta: {
    'semantic-ui-react/dropdown': {
      deps: ['semantic-ui-css/components/dropdown.css', 'semantic-ui-css/components/icon.css']
    }
  }
}
```

Right now the above code does not work _with non-global modules_.

DesignByOnyx:
This closes #1297

My editor is configured to use .editorconfig, and the indent_style is set to tab. As such, the `extension-cjs.js` file got completely rewritten (it used to use spaces).  If you want to see the diff without the whitespace, you can add `&w=1` to the URL above.


Gruntfile.js
```diff
@@ -142,6 +144,8 @@ module.exports = function (grunt) {
 					"src/system-extension-contextual.js",
 					"src/system-extension-script-module.js",
 					"src/system-extension-steal.js",
+					// "src/system-extension-module-loaded-twice.js",
```
DesignByOnyx: Matthew and I added this line here as it appeared to have been accidentally left out. However, this caused all sorts of issues during bundling with steal tools (tested on project). I left it commented on purpose so that someone else in the future doesn't try to put it back.

PR #1315, opened 2017-11-28
adds a npmignore file.

### LINKED ISSUE #1286: use npmignore
use the npmignore to exclude files we dont want to publish to npm like:
- test/
- docs/
- scripts/
and i think src/ as well because we are building your steal 

see https://github.com/npm/npm/wiki/Files-and-Ignores

the same applies to steal-tools

pYr0x:
ignore all unnecessary folders/files that users are not interested in when using npm.
close #1286

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,6 @@
+/test
+/scripts
+/src
+/.github
+Gruntfile.js
+.travis.yml


matthewp:
Thanks.

PR #1356, opened 2018-03-09
Setup eslint and fix lint issues in generated file

### LINKED ISSUE #1355: Fix lint setup

m-mujica:
Closes #1355

Summary:

- The original jshint setup was not being used at all
- The eslint command was only linting a specific rule
- The only rule I wasn't happy to turn off was `no-cond-assign` but that `while (assignment)` pattern is used in a few places inside the loader and I didn't want to change that
- Since steal has a gazillion files, I decided to run the lint only on steal-sans-promises.js (no point on linting the promise polyfill)
- The test script was changed a little bit to make sure steal files are compiled before the linter runs (and that the script that concatenates the files only runs once).
- @matthewp I swear I didn't sneak in any opinionated style rules in the eslint config :P 

💥 💥 💥


m-mujica:
of course tests failed.

PR #1359, opened 2018-03-14
Implements loader.rejectWithCodeFrame

matthewp:
This adds the `loader.rejectWithCodeFrame` (great name) function, to be
used with third party plugins that want to create better error messaging
where applicable. In DoneJS this will be used to provide better error
messages when rendering fails (like if the ViewModel throws).


matthewp:
This needs:

1. Only rewrite the stack trace to remove steal.js lines (like in commonjs).
2. Only add codeframing if the first line in the stack trace comes from the source module.

PR #1362, opened 2018-03-15
Improve the error message on mismatched package versions

### LINKED ISSUE #1346: Improve error message when there is a mismatch in package versions
During upgrades it's common to have wrong version numbers installed. If someone is requesting `package@3.0.0` but 4.0.0 is installed, we could give a good error message about this.

Probably we could even show the package.json where the wrong version number is listed.

matthewp:
This improves the error messaging and includes a link to documentation
about it. Closes #1346

### DIFF
--- a/ext/npm-crawl.js
+++ b/ext/npm-crawl.js
@@ -168,7 +168,7 @@ var crawl = {
 		}
 
 		var requestedVersion = childPkg.version;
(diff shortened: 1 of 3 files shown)


matthewp:
screen shot 2018-03-15 at 4 36 36 pm

matthewp:
Documented in https://github.com/stealjs/stealjs/pull/97

PR #1399, opened 2018-05-17
Move treeshake to translate hook

m-mujica:


src/extension-tree-shaking.js
```diff
@@ -97,16 +100,127 @@ addStealExtension(function (loader) {
...
+					]
+				}).code;
+			} catch (e) {
+				return Promise.reject(e);
+			}
+		});
+	}
+
+	var translate = loader.translate;
+	var es6RegEx = /(^\s*|[}\);\n]\s*)(import\s+(['"]|(\*\s+as\s+)?[^"'\(\)\n;]+\s+from\s+['"]|\{)|export\s+\*\s+from\s+["']|export\s+(\{|default|function|class|var|const|let|async\s+function))/;
```
m-mujica: I have to steal this from the es6 extension, `load.metadata.format` is not available at this point and I wasn't sure if there was a reliable way to do detection here other than using the regex.

m-mujica:
just force pushed some changes moving the treeshaking stuff to the translate hook; this way the transpile hook no longer deals with the treeshake plugin. This uncovered an issue with `reexecuteIfNecessary` calling [moduleSpecifierFromName](
https://github.com/stealjs/steal/blob/e2be1cdeca536faea93e666fff9da58a20bb269e/src/trace/trace.js#L57) before `load.metadata.dependencies` is set.

src/extension-tree-shaking.js
matthewp: Ah, yeah, maybe we can move this somewhere shared...

PR #1405, opened 2018-05-22
Replace a module while loading

### LINKED ISSUE #1403: Race condition in tree-shaking can cause module not to load
In the below scenario:

```
A (loads C)
B (loads C)
```

If B's dependency list happens after A, and we try to re-define C, we *could* remove C that is partially loaded.  I think the only way to fix this is to not use `loader.define` but rather other internal APIs. Investigating...

matthewp:
This makes it possible so that tree-shaking can replace a module when the graph is being loaded. Otherwise race conditions could leave a linkSet in a state that it will never resolve. Closes #1403


matthewp:
Fixed by #1403

PR #1414, opened 2018-05-31
Prevent properties called "require" from triggering the CJS dep detection

### LINKED ISSUE #1412: AMD CommonJS detection false positive
In lodash there's some code like:

```js
var types = freeModule && freeModule.require && freeModule.require('util').types;
```

It picks up 'util' as a dependency when it should not.

matthewp:
Closes #1412


matthewp:
All credit here goes to @James0x57

PR #1455, opened 2018-08-01
Backport getStealScript() from steal 2.x to fix Chrome 68 extension script issue

bmomberger-bitovi:
Chrome 68 is letting some extensions add their script tags to the DOM before synchronous scripts are run.  This breaks old Steal's expectation that the Steal script tag is the last script currently in the DOM.  This PR takes the more reliable getStealScript() function from the latest Steal and adapts it to 0.16 (no Electron check, otherwise the same).


matthewp:
👍 lgtm

PR #1472, opened 2018-10-26
Update config-bundlesPath.md

leoj3n:
> So, for example, if your script tag looks like:

This PR simply removes that sentence...

Also, I had trouble understanding what the documentation was saying about `bundlesPath` and made some edits to try and illustrate what I _think_ it is saying (and I'm probably wrong). If these docs are correct, they don't make sense to me. Maybe they could be made more clear somehow?

### DIFF
--- a/docs/config-bundlesPath.md
+++ b/docs/config-bundlesPath.md
@@ -7,7 +7,7 @@ bundles folder is located.
 @option {String} 
 
 A folder name that specifies the path to the production bundles.  By default,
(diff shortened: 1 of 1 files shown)


leoj3n:
The `bundlesPath` documentation is still confusing to me, but I'm going to close this non-critical issue even though I think the `bundlePath` docs could use some clarity improvements. Maybe give them a re-read sometime in the future to verify they're correct and see if they could be made more noob friendly.

PR #1479, opened 2018-11-27
Account for non-ES modules in parent tree

### LINKED ISSUE #1477: Tree-shaking extension fails for non-named imports
Assuming an imported file `helpers/helpers.js` that includes a handfull of imports like so:

```js
// helpers/helpers.js
import 'a.js'
import 'b.js'
```

In a stache file, use `can-import` to import the helpers file

```handlebars
// index.stache
<can-import from="~/helpers/" />
```

When loading that stache file the following error will occur.

screen shot 2018-11-21 at 9 06 58 am

This error occurs at https://github.com/stealjs/steal/blob/2488567bcb55a38958ba3877a280ad7202d3e28c/src/extension-tree-shaking.js#L106 because `parentImportNames` (`parentLoad.metadata.importNames`) is undefined.

A work around for this is to name the imports.

```js
// helpers/helpers.js
import A from 'a.js'
import B from 'b.js'
```

matthewp:
When re-walking the parent modules in order to perform tree-shaking,
	 account for the fact that there might be non-ES modules in the
	 tree.

Previously we were detecting non-ES modules in the "first pass" but not
latter passes that occur when more modules start being imported with
mutual dependencies.

Closes #1477


main.js
```diff
@@ -1043,9 +1043,15 @@ addStealExtension(function(loader) {
 				name
 			);
 
+			var parentIsESModule = parentLoad.metadata.format === "es6";
 			var parentImportNames = parentLoad.metadata.importNames;
 			var parentExportNames = parentLoad.metadata.exportNames;
 
+			// If this is an ES module then return true (indicating all are used)
```
mjstahl: Shouldn't this comment read 
> If this is NOT an ES module then return true (indicating all are used)

matthewp: Yeah good catch.

PR #1484, opened 2018-12-11
Allow CJS-style AMD to execute in require-order.

### LINKED ISSUE #1483: CJS-style AMD has problems with circular dependencies
This example demonstrates:

```js
define('one', [
    'require',
    'exports',
    'module',
    /*'two'*/
], function (require, exports, module) {
  module.exports = One;

  One.Two = require("./two");

  function One() {

  }
});

define('two', [
    'require',
    'exports',
    'module',
    /*'one'*/
], function (require, exports, module) {
  module.exports = Two;

  var One = require("./one");

  function Two() {

  }

  Two.prototype = Object.create(One.prototype);
});

define(['one']);
```

This code depends on the ability to define `module.exports` before a later function is called. This works fine in CommonJS but not in AMD-style cjs.

matthewp:
Traditionally AMD executes all dependencies before executing a module.
CommonJS, on the other hand, executes in require-order. That is, require
determines the execution order.

When transpiling CJS to AMD this can cause a problem where a module
written for CJS that contains a circular dependency where it depends on
require execution order can break.

This change allows for CJS-style AMD modules to execute in
require-order. There is a flag in Steal, `executingRequire` that
triggers this. We turn this flag on in a CJS-style AMD module, and only
execute dependencies up front, if those dependencies are listed before
the first `exports`, `module`, or `require` dependency.

Closes #1483


mjstahl:
I updated and the error below is what I got when using a `dev-bundle`:

screen shot 2018-12-12 at 9 57 45 am

I also tried a built version and got the following error:
screen shot 2018-12-12 at 10 00 31 am

PR #1485, opened 2018-12-12
Prevent tree-shaking the main

matthewp:
The main module should not be tree-shaken. We can detect this because
its parents === 0.


matthewp:
Fixes #1486

Issue #1463, opened 2018-09-18
If a package is missing, warn which file imported it.


justinbmeyer:
![image](https://user-images.githubusercontent.com/78602/45692984-addc2200-bb21-11e8-9e02-caed1e352857.png)

The above warning should be visible by using this commit: https://github.com/canjs/can-stache-route-helpers/commit/5085b6581994ff818871144daa998c1f3ef1ae48

of `can-stache-route-helpers`

justinbmeyer:
To make this warn now, you'll probably manually have to delete `can-define` from your `node_modules` folder as the bug in `can-define` is fixed.  The warning might be because the module importing `can-define` was within `node_modules`.

leoj3n:
I have the same issue with the `events` package. I can't figure out who is trying to use `events` when it 404s. All I know is that it 404'd. Even if I manually bundle `events`:

```json
    "map": {
      "events": "~/node_modules/events/events"
    },
    "bundle": [
      "~/node_modules/events/events",
```

Which causes `dist/bundles/my-project/node_modules/events/events.js` to exist, it still 404s because it is looking for `/dist/node_modules/events/events.js` which is outside the `my-project` bundle which is probably related to what @justinbmeyer mentioned:

> The warning might be because the module importing `can-define` was within `node_modules`.

Is there some way to create a mapping to fix this? Currently the only solution I have is to manually create the `/dist/node_modules/events/` directory and copy `events.js` as part of the build step.

cherifGsoul:
My solution to solve this issue is to add ```stack``` property to https://github.com/stealjs/steal/blob/8c81c5ac3ffd4d912f95b24bc091236af459db5a/ext/npm-extension.js#L400

```newError.stack = error.stack```

cherifGsoul:
After discussion with @matthewp the error will look like the following screenshot:

<img width="1425" alt="screen shot 2019-01-11 at 8 48 04 pm" src="https://user-images.githubusercontent.com/109013/51056149-5ab34100-15e2-11e9-9e50-f30faecc3e3d.png">

PR #1493, opened 2019-01-11
warn with the importer in case of missing module

cherifGsoul:
This adds the importer file of a missing module to the error message by prepending the stack to error.

Fixes https://github.com/stealjs/steal/issues/1463

### DIFF
--- a/ext/npm-extension.js
+++ b/ext/npm-extension.js
@@ -393,11 +393,14 @@ exports.addExtension = function addNpmExtension(System){
 			if ((statusCode === 404 || statusCode === 0) &&
 				utils.moduleName.isBareIdentifier(load.name) &&
 				!utils.pkg.isRoot(loader, load.metadata.npmPackage)) {
-				var newError = new Error([
-					"Could not load '" + load.name + "'",
+				var errorMsg = ["Could not load '" + load.name + "'",
 					"Is this an npm module not saved in your package.json?"
-				].join("\n"));
+				].join("\n");
+				var newError = new Error();
 				newError.statusCode = error.statusCode;
+				newError.stack = newError.stack + error.stack;
+				newError.message = errorMsg;
+				
 				throw newError;
 			} else {
 				throw error;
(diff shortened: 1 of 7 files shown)


ext/npm-extension.js
```diff
@@ -393,11 +393,14 @@ exports.addExtension = function addNpmExtension(System){
 			if ((statusCode === 404 || statusCode === 0) &&
 				utils.moduleName.isBareIdentifier(load.name) &&
 				!utils.pkg.isRoot(loader, load.metadata.npmPackage)) {
-				var newError = new Error([
-					"Could not load '" + load.name + "'",
+				var errorMsg = ["Could not load '" + load.name + "'",
 					"Is this an npm module not saved in your package.json?"
-				].join("\n"));
+				].join("\n");
+				var newError = new Error();
```
matthewp: I think you want to pass errorMsg into the constructor, otherwise it will not be part of the current `.stack`, so prepending won't add this part.

matthewp:
Looks good now 👍

test/npm_nested_import_errors/main.js
```diff
@@ -0,0 +1,2 @@
+//require("dep");
```
matthewp: dep is not being imported so the error won't occur, right?

matthewp:
I think the test is not working. I don't think the "dep" is being imported so the error does not occur. This is what is happening in CI:

screen shot 2019-01-14 at 5 59 54 pm

test/npm_nested_import_errors/main.js
cherifGsoul: yes, I imported here ```System.import('dep').then(null, function(err){``` line 15 of dev.html

cherifGsoul:
@matthewp I fixed the test, the ```dep``` file was inside ```node_modules``` which is ignored, can take a look please (tomorrow).

matthewp:
Looks good!

PR #1501, opened 2019-07-31
Use a stricter RegExp to detect System.register modules

### LINKED ISSUE #1500: Steal gets confused when script contains *System.register*

- [X] Always
- [ ] Sometimes
- [ ] Rarely
- [ ] Unable
- [ ] I didn’t try

__Description:__

When a script contains the text **System.register** then StealJS seams to assume that this file is a SystemJS module. That would be fine if the file actually did call **System.register()** but actually my script calls **FileSystem.registerFileSystem()** which has nothing to do with SystemJS. Looks like the regular expression which looks for `System.register` is not exact enough. Maybe it could simply be changed to `/\bSystem\.register/`?

__Steps to reproduce:__

Try to load a file with the following content:

```javascript
require("fs.js");
FileSystem.registerFileSystem();
```

__Expected results:__

Steal should load the dependency `fs.js` (or complain that this file could not be loaded).

__Actual results:__

The error `ReferenceError: require is not defined` is thrown: I think this is because StealJS assumes the file is a SystemJS module instead of a CommonJS module and therefor doesn't provide the require function.

Remove the second line in the example or change it so it doesn't match **System.register** and then it works correctly.

__Environment:__

| Steal version              | 2.2.1
| Steal-tools version    | -
| node -v                       | -
| npm -v                        | -
| Browser                      | Chrome 75.0.3770.100
| Operating system      | Debian Linux 10

m-mujica:
Fixes #1500


matthewp:
Tests failing.

giphy

m-mujica:
Ok, so initially the tests failed because of an old appium-version.

I fixed that, but now I get an actual test failure on the iphone 7. 

0002screenshot

I'm going to try to debug this, and in the meantime I'll re-run the tests to make sure the issue is consistent.

m-mujica:
Ended up skipping the tests for safari, it didn't seem like a simple fix... Created an issue to tackle the stack trace in Safari later.

PR #1503, opened 2019-08-01
Fix security vulnerability introduced by eslint<4.18.2

m-mujica:
Ran into some issues with the files lookup, it seems passing only the directories is better than globbing. I intentionally added a param reassign to make sure eslint was indeed working.

```
npm run lint

> steal@2.2.2 lint /Users/mmujica/Projects/steal
> eslint src ext


/Users/mmujica/Projects/steal/src/json/json.js
  4:2  error  Assignment to function parameter 'loader'  no-param-reassign

✖ 1 problem (1 error, 0 warnings)
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "scripts": {
     "build": "grunt",
-    "lint": "eslint {src,ext}/**/*.js --quiet",
+    "lint": "eslint src ext",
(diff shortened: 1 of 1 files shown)


m-mujica:
I'm sure there is no issues with me merge this PR 💅

PR #1513, opened 2019-11-05
All CommonJS and AMD modules to be transpiled

### LINKED ISSUE #1276: Allow transpiling module formats other than ESM
__tldr;__ Currently only ES Modules (modules containing the `import` or `export` keyword) are transpiled through Babel. Users of other formats, like CommonJS, would like to also be able to transpile their code for production.

# The Problem

Users in legacy codebases

matthewp:
This enables AMD and CommonJS modules to also be transpiled (run through Babel). Still a little bit of a WIP. Closes #1276


main.js
```diff
@@ -692,10 +692,11 @@ addStealExtension(function addMetaDeps(loader) {
 	}
 
 	loader.transpile = function (load) {
+		// TODO this needs to change
```
matthewp: Now that `.transpile()` is called on all formats, prependDeps needs to check the load.format.

main.js
```diff
@@ -1735,6 +1761,7 @@ addStealExtension(function addCacheBust(loader) {
 	System.ext = Object.create(null);
 	System.logLevel = 0;
 	System.forceES5 = true;
+	System.transpileAllFormats = true;
```
matthewp: Not sure if this is the right option. In the PR we said `transpile` would be it.

Also this has to default to false or it will be a breaking change.

matthewp:
There are a few issues with this PR but it basically works. Also need to test against steal-tools as I'm not sure what the implications this might have on that.

main.js
matthewp: `steal.loader.transpile` is a function that's used to do transpilation. We could support the `transpile: true` option, but this would be tricky to implement. I don't think a slightly better name is actually worth it and `transpileAllFormats` is just fine.

matthewp:
Need a way to test this. My thought:

Have a custom babel plugin. Maybe it could change a function signature in some way. We would know that it works by calling same function. For example:

```js
module.exports = function() {
  return 1;
};
```

Transpiles to:

```js
module.exports = function() {
  return 2;
};
```

But only for a certain module. Seems reasonable.

